PR #189, opened 2025-01-03
Wss test client

burundiocibu:
Client connects to the server but it doesn't seem to receive data from the the connection. Back to trying to capture and decode a wpa negotiation from my phone to the server...


scripts/ble_test.py
```diff
@@ -0,0 +1,33 @@
...
+    print(data)
+
+
+async def main():
+    cfg = configparser.ConfigParser()
+    cfg.read(str(Path.home() / ".pitboss"))
+    model = cfg["pitboss"]["model"]
+    device_address = cfg["pitboss"]["device_address"]
+
+    ble_device = await BleakScanner.find_device_by_address(device_address)
```
dknowles2, line 24: I think the address may be unstable. You're probably be better off using `BleakScanner.find_device_by_name()` and look it up based on the `grill_id` (the Bluetooth broadcast name is the same as the grill id)

If you did that, you could combine both of these into a single script, choosing the transport based on an optparse flag or something.

I'd also call the script "debug" instead of having it end with "_test" (since that implies that these may be unit tests, which they're not)

dknowles2 (changes requested):

dknowles2:
> Client connects to the server but it doesn't seem to receive data from the the connection. 

My connection was super flaky when I was testing this out earlier in the week. Even the iOS client was dropping the connection constantly. I couldn't tell if it was something screwed up with my account or if the PitBoss servers are just being wonky.

> Back to trying to capture and decode a wpa negotiation from my phone to the server...

I've been using https://httptoolkit.com/ with an Android emulator to sniff traffic. There's a few different sites that publish APKs that you can side-load into the emulated device, and HTTP Toolkit is able to circumvent certificate pinning.

scripts/ble_test.py
burundiocibu: Great ideas all. I'll get on it.

burundiocibu:
BTW, I see the same flakiness from my IOS client as well; even when I've not tried connecting to the web socket via other means for days. And that could be on either the pitboss server sid or in the grill<-> cloud connection being formed.

Thanks for the pointer on the httptoolkit not one I've used but it looks powerful.

scripts/ble_test.py
burundiocibu: I need to attach the dev container to the host's bluetooth hardware or require that the new "debug_transport.py" script be run outside a dev container. I'm going with the later to keep the dev container hardware agnostic. Let me know if you'd prefer otherwise.

dknowles2: Does the devcontainer not allow bluetooth? I thought it did--maybe I broke it at some point.

Maybe try merging this into the devcontainer config?

```
    "runArgs": [
        "--group-add",
        "bluetooth"
    ],
    "features": {
        "ghcr.io/devcontainers-extra/features/apt-packages:1": {
            "packages": [
                "bluez",
                "libpcap-dev",
                "libudev-dev"
            ]
        }
    }

```

burundiocibu: I've made suggested changes and have tested it on bare metal (a pi5) that is literally 1 meter in front of a running grill  and the BLE find would never find. Sigh. I blame this on general wonky-ness of the pitboss sw.

I've also add the bluez (back) into the container and am now figuring out the "right" way to connect the physical device to the container in this type of workflow. So I marked this as draft. Shoulda don't that at the start.

BTW. I'm impressed w your use of the dev container workflow/vscode/gitlab ci. I've learned a few tricks working with this project; thanks. FWIW, my typical dev container is running on a ubuntu LTS box. The bluetooth device is the first bit of friction that might indicate a difference in our dev container's host. Well done.

burundiocibu:
This is able as far as I can go without a more stable BLE from the grill. I did get the bluetooth stack going in the container.

scripts/ble_test.py
dknowles2, line 1: Do we still need this file?

scripts/debug_transport.py
```diff
@@ -0,0 +1,76 @@
...
+
+    grill_models = [grill.name for grill in grills.get_grills()]
+    if model not in grill_models:
+        print(f"Invalid grill model: {model}")
+        print(grill_models)
+        return
+
+    try:
+        print(f"Using BLE to find device: {model}")
+        ble_device = await BleakScanner.find_device_by_name(model)
```
dknowles2, line 34: I thought the Bluetooth name was the same as the device id? I don't think you'll find anything looking for a model number.

scripts/debug_transport.py
```diff
@@ -0,0 +1,76 @@
...
+        boss = pb.PitBoss(pb.BleConnection(ble_device), model)
+        await boss.subscribe_state(state_callback)
+        await boss.start()
+        while True:
+            asyncio.sleep(0.1)
+            if boss.is_connected():
+                print(boss)
+            break
+    except FileNotFoundError as ex:
```
dknowles2, line 43: This `try/except` block is too broad--it should only cover the `BleakScanner.find_device_by_name()` call. There's no way that anything else would ever raise a `FileNotFoundError`.

scripts/debug_transport.py
```diff
@@ -0,0 +1,76 @@
...
+    grill_models = [grill.name for grill in grills.get_grills()]
+    if model not in grill_models:
+        print(f"Invalid grill model: {model}")
+        print(grill_models)
+        return
+
+    try:
+        print(f"Using BLE to find device: {model}")
+        ble_device = await BleakScanner.find_device_by_name(model)
+        boss = pb.PitBoss(pb.BleConnection(ble_device), model)
```
dknowles2, line 35: The `PitBoss` creation and loop is generic to both transport types. It only needs to be created once.

scripts/debug_transport.py
```diff
@@ -0,0 +1,76 @@
...
+    model = cfg["pitboss"]["model"]
+
+    grill_models = [grill.name for grill in grills.get_grills()]
+    if model not in grill_models:
+        print(f"Invalid grill model: {model}")
+        print(grill_models)
+        return
+
+    try:
+        print(f"Using BLE to find device: {model}")
```
dknowles2, line 33: I'd set up an [argparse](https://docs.python.org/3/library/argparse.html) flag that specifies what transport you want to use. Something like this:

```py
from argparse import ArgumentParser
parser = ArgumentParser()
parser.add_argument('transport', metavar='transport', type=str, choices=('bluetooth', 'websocket'), nargs=2)

...

async def main():
  args = parser.parse_args()
  ...
  if args.transport == 'bluetooth':
    ble_device = find_device_by_name(...)
    transport = pb.BleConnection(ble_device)
  else:
    transport = pb.WebSocketConnection(...)

  boss = PitBoss(transport, model)
  ...
```

You could even make `model` and `device_id` flags as well so that you don't need to read from a `.pitboss` file.

.devcontainer.json
```diff
@@ -33,12 +33,24 @@
...
     "runArgs": [
+        // net=host seems to be required for the bluez service to find the bluetooth adaptor
+        "--net=host",
         "--group-add",
-        "dialout"
+        "dialout",
+        "--group-add",
+        "bluetooth"
     ],
+    "postStartCommand": "sudo service dbus start && sudo service bluetooth start",
```
dknowles2, line 43: I'll have to poke at it more when I have time, but I know the HomeAssistant devcontainer can pass-thru the bluetooth controller and it seems to work fine. I've never had to stop the bluetooth service on my laptop when developing integrations.

scripts/ble_test.py
burundiocibu: nope. now deleted.

scripts/debug_transport.py
burundiocibu: This was just me misunderstanding what you had said about the device name. Correcting.

scripts/debug_transport.py
burundiocibu: So my thought was to try to test both BLE and wss in a single run. Create a BLE connected pitboss object, use that to get the grill_id (which from above is the way to do it), then use that grill_id to form the web socket.

I'll do what you suggest below w cmd line args to debug each transport separately.

.devcontainer.json
burundiocibu: Yeah, the ha dev container behaviour is beyond me at the moment. I've refactored the debug wrapper but need to check to see how it responds using the grill_id as opposed to the model. I've got more testing to do...

dknowles2 (commented):
### Code Review Summary

1. **Bug in `scripts/debug_transport.py`**:
   - On line 63, `asyncio.sleep(0.1)` is called without `await`. It needs to be `await asyncio.sleep(0.1)`. Without `await`, it creates an un-awaited coroutine in an infinite tight CPU spin loop.
   - `BleakScanner.find_device_by_name()` can return `None` if no matching device is found. Adding a check `if ble_device is None:` before passing it to `pb.BleConnection(ble_device)` will prevent a cryptic failure down the line.

2. **Formatting in `pytboss/__init__.py`**:
   - Line 5 has `# noqa F401` (missing colon), whereas lines 3 and 4 use `# noqa: F401`.

3. **Devcontainer / Docs**:
   - The added devcontainer configuration (`bluez`, `libpcap-dev`, `libudev-dev`, `--net=host`, `bluetooth` group) and updated README instructions for bluetooth setup look good and make BLE development inside devcontainers easier.

4. **WSS Status**:
   - Noted that this PR is currently in Draft status while WSS message reception/decoding is being worked on.

dknowles2 (commented):
### Code Review Summary

**Verdict: Flag Concerns**

#### 1. Bugs in `scripts/debug_transport.py`:
- **Un-awaited coroutine (Line 63):** `asyncio.sleep(0.1)` is called without `await` inside `while True:`. This results in a CPU-bound infinite spin loop and a `RuntimeWarning: coroutine 'sleep' was never awaited`. It should be `await asyncio.sleep(0.1)`.
- **Missing None check for BLE device:** `BleakScanner.find_device_by_name()` can return `None` if no device matches `args.grill_id`. Explicitly checking `if ble_device is None:` before creating `BleConnection(ble_device)` provides a much clearer error message.
- **Transport option handling:** `--transport` does not constrain allowed values. Passing an invalid value silently falls back to `WebSocketConnection`. Adding `choices=["ble", "ws"]` to `add_argument` would avoid unexpected behavior.

#### 2. Flake8/Ruff Directive in `pytboss/__init__.py`:
- The comment `# noqa F401` is missing the colon (`# noqa: F401`).

#### 3. Merge State & PR Status:
- PR #189 is in **Draft** state and has merge conflicts with `main` (DIRTY state).

David (`dknowles2`) has final say on PR approval and merge.

Issue #228, opened 2025-04-17
Add support for 850DX


airdamien:
Increasing the range in the dump_grills.py gets the 850DX in the list, I've shuffled that over to replace the grills. json 

def main():
    cfg = configparser.ConfigParser()
    cfg.read(str(Path.home() / ".pitboss"))
    auth = login(cfg["pitboss"]["username"], cfg["pitboss"]["password"])
    grills = {}
    for i in range(1, 111):

      "created_at": "2024-05-20T20:49:31.000000Z",
      "deleted_at": null,
      "id": 16,
      "name": "PBL2",
      "platform_id": 3,
      "site_id": 1,

added PBL2 
logging.basicConfig(level=logging.DEBUG)  # Log all HTTP requests to stderr.
API_URL = "https://api-prod.dansonscorp.com/api/v1"
CONTROL_BOARDS = (
    "LBL",
    "LFS",
    "PBA",
    "PBC",
    "PBG",
    "PBL",
    "PBM",
    "PBP",
    "PBT",
    "PBV",
    "PBL2",
)

import asyncio
from bleak import BleakScanner
from pytboss import BleConnection, PitBoss

async def state_callback(data):
    print(data)

async def main():
    ble_device = await BleakScanner.find_device_by_address("0B:BC:CA:68:D3:0A")
    #ble_device = await BleakScanner.find_device_by_address("00:63:88:A2:5D:48")
    model = "PB850DX"  # Or your model. See below.
    boss = PitBoss(BleConnection(ble_device), model)
    # Subscribe to updates from the smoker.
    await boss.subscribe_state(state_callback)
    await boss.start()
    while True:
        asyncio.sleep(0.1)

asyncio.run(main())

But my copy of the example does not find the device, I'd really like to get this working for homeassistant. 
appreciated

airdamien:
Ok, so some progress, I'm not getting any data that's triggering the state received or data, and I'm not sure how or if I need to fix the 
/pytboss/./cb.py:28: RuntimeWarning: coroutine 'sleep' was never awaited
  asyncio.sleep(0.1)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback



 ➜  pytboss git:(main) ✗ python3 ./cb.py
WAITING FOR DEVICE
6A9215A4-6ABA-REDACTES: PBL2-REDACTED
GETTING CONNECTION
init
[...]

PR #230, opened 2025-04-21
Update grill definitions

dknowles2:
This addresses #228 and will unblock https://github.com/dknowles2/ha-pitboss/issues/102

### DIFF
--- a/pytboss/grills.json
+++ b/pytboss/grills.json
@@ -1,141 +1,4 @@
 {
-  "AustinXL Wi-Fi Upgrade": {
-    "category": null,
-    "control_board": {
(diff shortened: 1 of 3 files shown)


airdamien:
@dknowles2 I have this working, I have cracked the password problem, I have it working in HA. I don't have permissions to commit back my branch.   

[diffs.tar.gz](https://github.com/user-attachments/files/19840118/diffs.tar.gz)

airdamien:
you'll also need to support PBL2- for HA, I think that was about the only real change in the ha side, other than my rewiring it to use a local lib and not system.

PR #232, opened 2025-04-22
Proper password handling 

airdamien:
I can only test on the PBL2- 850DX.. but this is working great for me so far manual and with homeassistant.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -44,7 +44,6 @@ class PitBoss:
         self.fs = FileSystem(conn)
         self.config = Config(conn)
         self.spec: Grill = get_grill(grill_model)
-        #print(get_grill(grill_model))
         self._conn = conn
         self._conn.set_state_callback(self._on_state_received)
         self._conn.set_vdata_callback(self._on_vdata_received)
@@ -53,13 +52,9 @@ class PitBoss:
         self._state_callbacks: list[StateCallback] = []
         self._vdata_callbacks: list[VDataCallback] = []
         self._state = StateDict()
-        #print("password")
-        #print(password)
 
     def is_connected(self) -> bool:
         """Returns whether we are actively connected to the grill."""
-        #print("is_connected")
-        #print(self._conn.is_connected())
         return self._conn.is_connected()
 
     async def start(self) -> None:
@@ -68,11 +63,9 @@ class PitBoss:
         Required to be called before the API can be used.
         """
         await self._conn.connect()
-        #print("start connect await done")
 
     async def stop(self) -> None:
         """Stops any background polling."""
-        #print("STOP")
         await self._conn.disconnect()
 
     async def subscribe_state(self, callback: StateCallback):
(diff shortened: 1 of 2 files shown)


airdamien:
@dknowles2 is this a F to C problem? 
 ```
E                   AssertionError: p1Target: 88 != 191
E                   assert 88 == 191

tests/test_grills.py:183: AssertionError
```

**Comment by airdamien** [comment]
nevermind, I figured it out

**Comment by airdamien** [comment]
now pytest is sticking on test_wss, I haven't touched that, suspect it's either my environment or not a me-issue

**Comment by airdamien** [comment]
pytest api is still failing, it's the await_once, I think that for the password logic to work, really need the nested awaits as there's a big time component that has to be correct to generate the acceptable psw.

**Comment by dknowles2** [comment]
> @dknowles2 is this a F to C problem?
> 
> ```
> E                   AssertionError: p1Target: 88 != 191
> E                   assert 88 == 191
> 
> tests/test_grills.py:183: AssertionError
> ```

I'll address this in my PR that updates the grill definitions once I get some time to look at it.

**Comment by dknowles2** [comment]
> pytest api is still failing, it's the await_once, I think that for the password logic to work, really need the nested awaits as there's a big time component that has to be correct to generate the acceptable psw.

The problem is that the expectations no longer match. Calls that need a password now need to send a `PB.GetTime` RPC which isn't accounted for in the mocks.

**Comment by airdamien** [comment]
> > pytest api is still failing, it's the await_once, I think that for the password logic to work, really need the nested awaits as there's a big time component that has to be correct to generate the acceptable psw.
> 
> The problem is that the expectations no longer match. Calls that need a password now need to send a `PB.GetTime` RPC which isn't accounted for in the mocks.

yeah, do you want me to implement that? Unsure if the assert_once was purposeful.

File: pytboss/grills.json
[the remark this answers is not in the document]
Comment (opn-cbutkus) [reply pytboss/grills.json:1]: just reverted the grills.json and pushed.

**Review by opn-cbutkus — APPROVED** [review]

File: pytboss/api.py [hunk 1]
```diff
@@ -128,14 +133,112 @@ async def _on_vdata_received(self, payload: str):
...
-        if self._password:
-            params["psw"] = self._password.encode("utf-8").hex()
-        return params
+    async def _get_codec_time(self) -> int:
+        """Port of getCodecTime() from JavaScript using actual grill time."""
+        try:
+            response = await self.get_time()
+            uptime = response.get("time", 0)
+            return int(max(uptime - 5, 0) / 10)
+        except (RPCError, GrillUnavailable, NotConnectedError) as e:
```
dknowles2, line 142: You can just catch `pytboss.exceptions.Error` which covers all of these.

pytboss/api.py
```diff
@@ -128,14 +133,112 @@ async def _on_vdata_received(self, payload: str):
...
+            # Fallback to local time if we can't get grill time
+            _LOGGER.debug("Failed to get grill time, using local time: %s", e)
+            return int(max(time.time() - 5, 0) / 10)
+
+    def _get_codec_key(self, key_bytes: list[int], time_val: int) -> list[int]:
```
dknowles2, line 147: Let's move `_get_codec_key()` and `_codec()` to a new module called `auth`. And then pull the loop out of the `_authenticate()` method into a (public) function called `encrypt_password()`. That way the encoding routine is testable on its own without relying on private methods.

dknowles2 (changes requested):

Issue #167, opened 2024-12-15
Support for OTA ?


IDerr:
Hello,

do you plan to add some more rpc calls in this library, I have found those but all doesn't seem to be part of this library ?
```
PB.GetState
PB.GetVirtualData
PB.SetDevicePassword
FS.Get
FS.Put
PBL.LoadFirmware
PBL.LoadFirmwareStatus
PB.GetTime
PBL.GetLoaderVersion
PB.GetFirmwareVersion
FS.Rename
RPC.List
PBL.CopyFile
FS.Remove
PB.SendMCUCommand
PB.SetWifiCredentials
Config.Set
Config.Get
PBL.GetWifiScanStatus
Wifi.Scan
PBL.StartWifiScan
Sys.Reboot
Config.Save
PB.SetWiFiUpdateFrequency
Sys.GetInfo
RPC.Ping
FS.List
OTA.Start
OTA.Status
PB.LoadMCUFirmware
PB.LoadMCUFirmwareStatus
PB.QueryMCUFirmwareVersion
PB.SetVirtualData
PB.RenameDevice
PB.WiFiAwakeWDT
```

trailro:
Went through the list against a real grill — a PB1600PS1 on 0.5.7 — and against the vendor firmware images. Where things stand, and what I think is worth doing with each.

**Of the 35 listed, 20 are wrapped today.** Six of the ones you marked as missing have landed since you filed this: `PB.SetDevicePassword`, `PB.GetTime`, `PB.WiFiAwakeWDT`, `RPC.Ping`, `Sys.Reboot` and `Sys.GetInfo` (the last four via #530).

One correction to the list itself: `PB.SetWiFiUpdateFrequency` was marked OK, and the wrapper existed — but it asked for `PB.SetWifiUpdateFrequency`, which every firmware answers with "name not found". Fixed in #559.

## Worth wrapping — done in #561

**`RPC.List`, `PBL.GetLoaderVersion`.** Reads, no arguments, no password, cannot change anything. `RPC.List` is the one with leverage: what a grill answers is not uniform, and both #543 and #458 currently turn on that question. `PBL.GetLoaderVersion` reports a version that is genuinely separate from the firmware one — the grill above answers `0.5.7` for firmware and `0.2.2` for the loader.

## Worth wrapping — not done here

[...]

PR #458, opened 2026-06-26
Add OTA update support

dknowles2:
Adds Mongoose OS's OTA service — `OTA.Start` and `OTA.Status` — as `PitBoss.ota`, alongside the existing `fs` and `config` helpers.

Part of #167.

## Narrowed from the original branch

This branch previously added nine loose RPC wrappers to `PitBoss` as well. Those are gone; what remains is `ota.py`, its tests, and the three lines that wire it up.

Two of the nine have since landed anyway, via #530:

| was | now |
|---|---|
| `reboot()` | merged in #530, unchanged |
| `wifi_awake_wdt()` | merged in #530 as **`request_fast_updates()`** |

The rename matters and is worth not undoing here. `PB.WiFiAwakeWDT` does not keep the WiFi module awake — it sets a five-minute countdown during which the status-push timer reschedules to `wsFastInterval` instead of `wsSlowInterval` (5 s instead of 60 s), and it only affects the cloud WebSocket, only while the grill is on. The old name described something the RPC does not do.

The remaining seven — `list_rpcs`, `rename_device`, `set_wifi_credentials`, `load_mcu_firmware`, `get_mcu_firmware_load_status`, `query_mcu_firmware_version`, `scan_wifi` — are not OTA and are dropped rather than carried. Each is easier to argue for on its own; several are ones where a wrong call takes the grill off its network or leaves the MCU mid-flash.

**One judgement call worth naming:** the three `PB.*MCUFirmware*` wrappers are firmware-loading, so they are arguably in scope for "OTA". I read the instruction strictly — `OTA.*` is Mongoose's own service and `PB.LoadMCUFirmware` is a different mechanism against a different processor. Easy to add back if that reading is wrong.

## Auth

No `_authenticate` call, deliberately. `OTA.*` is a Mongoose core namespace, not `PB.*`, and `checkPassword` only guards the handlers the app registers — the same reason `fs.py` and `config.py` do not authenticate either. Consistent with the rest of the library.

## What is not verified

`OTA.Start` and `OTA.Status` are **not modelled in `fake_firmware/init.js`**, because it only registers the app's own handlers — the same gap `Sys.Reboot` has. So the wrappers match [Mongoose's documented RPC service](https://mongoose-os.com/docs/mongoose-os/api/rpc/rpc-service-ota.md), and the tests pin the request shape, but nothing here demonstrates a real grill answering them. Flagging rather than implying otherwise.

Anyone testing this against hardware should also know it is the sharpest edge in the library: a bad OTA image is how a grill stops working entirely. `commit_timeout` exists for exactly that and is worth using.

## Testing

745 tests pass (3 new), `ruff check`, `ruff format --check` and `mypy .` all clean.

### DIFF
--- a/pytboss/__init__.py
+++ b/pytboss/__init__.py
@@ -2,4 +2,5 @@
 
 from .api import PitBoss  # noqa: F401
 from .ble import BleConnection  # noqa: F401
+from .ota import OTA  # noqa: F401
 from .wss import WebSocketConnection  # noqa: F401
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -14,6 +14,7 @@ from .config import Config
 from .exceptions import UnsupportedOperation
 from .fs import FileSystem
 from .grills import Grill, StateDict, get_grill
+from .ota import OTA
 from .transport import RPCResult, Transport, as_dict
 
 _UPTIME_TTL = 60.0
@@ -74,6 +75,9 @@ class PitBoss:
     config: Config
     """Configuration operations."""
 
+    ota: OTA
+    """Over-the-air firmware update operations."""
+
     def __init__(
         self,
         conn: Transport,
@@ -96,6 +100,7 @@ class PitBoss:
         """
         self.fs = FileSystem(conn)
         self.config = Config(conn)
+        self.ota = OTA(conn)
         self._grill_model = grill_model
         self._control_board = control_board
         self._conn = conn
--- a/pytboss/ota.py
+++ b/pytboss/ota.py
@@ -1,0 +1,46 @@
+"""Client library for Mongoose OS OTA (Over-The-Air update) RPCs."""
+
+from .transport import RPCResult, Transport, as_dict
+
+
+class OTA:
+    """Client library for Mongoose OS OTA RPCs.
+
+    Also see: https://mongoose-os.com/docs/mongoose-os/api/rpc/rpc-service-ota.md
+    """
+
+    def __init__(self, conn: Transport) -> None:
+        """Initializes the class.
+
+        :param conn: Transport for the device.
+        """
+        self._conn = conn
+
+    async def start(self, url: str, commit_timeout: int | None = None) -> RPCResult:
+        """Initiates an OTA update.
+
+        Returns whatever the device answers rather than promising a shape:
+        `OTA.*` is a Mongoose core service, so it is not modelled in
+        `fake_firmware/init.js` and nothing here has been seen answering a
+        real grill.
+
+        :param url: URL of the firmware zip file to download.
+        :param commit_timeout: Optional timeout (in seconds) to auto-commit the
+            new firmware. If the device is not committed within this time it
+            rolls back automatically.
+        """
+        params: dict = {"url": url}
+        if commit_timeout is not None:
+            params["commit_timeout"] = commit_timeout
+        return await self._conn.send_command("OTA.Start", params)
+
+    async def get_status(self) -> dict:
+        """Returns the current OTA update status.
+
+        The returned dict typically contains:
+          - ``state``: current OTA state string (e.g. ``"idle"``, ``"progress"``,
+            ``"error"``)
+          - ``msg``: human-readable message
+          - ``progress``: download progress (0–100) while downloading
+        """
+        return as_dict(await self._conn.send_command("OTA.Status", {}))
--- a/tests/test_ota.py
+++ b/tests/test_ota.py
@@ -1,0 +1,44 @@
+"""Tests for OTA RPC client."""
+
+from unittest import mock
+
+import pytest
+
+from pytboss.ota import OTA
+
+
+@pytest.fixture
+def mock_conn():
+    conn = mock.AsyncMock()
+    conn.send_command = mock.AsyncMock(return_value={})
+    return conn
+
+
+async def test_ota_start(mock_conn):
+    ota = OTA(mock_conn)
+    await ota.start("http://example.com/firmware.zip")
+    mock_conn.send_command.assert_awaited_once_with(
+        "OTA.Start", {"url": "http://example.com/firmware.zip"}
+    )
+
+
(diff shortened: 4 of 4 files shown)


trailro:
Flagging a downstream use, in case it helps prioritise this one.

Two entities in ha-pitboss are waiting on `reboot()` and `wifi_awake_wdt()` — restarting the WiFi module is the standard fix when it wedges, and the keep-awake watchdog is what stops it dropping the connection. Neither is reachable today without `_conn.send_command`, and that integration uses no private pytboss APIs anywhere, so I would rather not be the first.

I split my findings out of dknowles2/pytboss#506 into separate issues just now and deliberately did **not** file one for these two, since this PR already implements them. Happy to review or rebase it if that is useful.

trailro:
Took the `OTA.*` names to a grill, and the wrappers do not match what the firmware registers. Read-only throughout — `RPC.List` and `RPC.Describe`; I have not run an actual update.

## `OTA.Start` is not on this line

`RPC.List` on a PB1600PS1 (Mongoose OS, `mg_version` 2.17.0) returns `OTA.Begin`, `OTA.Write`, `OTA.End`, `OTA.Update`, `OTA.Commit`, `OTA.Revert`, `OTA.CreateSnapshot`, `OTA.GetBootState`, `OTA.SetBootState` and `OTA.Status` — and **no `OTA.Start`**. The method that does what `start()` means is `OTA.Update`:

```
RPC.Describe {"name": "OTA.Update"}
-> {"name": "OTA.Update",
    "args_fmt": "{file: %Q, url: %Q, timeout: %d, commit_timeout: %d, ignore_same_version: %B}"}
```

`url` and `commit_timeout` are there, exactly as the PR sends them — and [the docs this PR cites](https://mongoose-os.com/docs/mongoose-os/api/rpc/rpc-service-ota.md) document that same pair under `OTA.Update`, with no `OTA.Start` anywhere on the page. So the parameters are right and the method name is not.

## `OTA.Start` is real, on the other firmware line

It is the ESP-IDF firmware's own name — present in every image from 16.2.0 to 16.8.3, alongside `OTA.BluetoothStart`/`OTA.BluetoothFinalize`, and **without** `OTA.Update`/`Begin`/`Write`/`End`/`Commit`/`Revert`/`CreateSnapshot`. Precisely the mirror image of the Mongoose set. So this PR is correct for those grills and inapplicable to Mongoose ones, which is the same family split as on #543.

Worth ruling out one hypothesis explicitly: this is **not** an app-version thing. The mJS apps register zero `OTA.*` handlers in every version from 0.2.3 to 0.6.0 — the OTA surface comes from the Mongoose base, which Dansons does not update (ours reports `fw_id` 20201001, and app self-update has been disabled since 0.5.7). There is no 0.x.x where `OTA.Start` exists.

## `OTA.Status` returns a different shape

Documented in the PR as `state` (string: `"idle"`/`"progress"`/`"error"`), `msg`, `progress`. Actual reply:

```json
{"state": 0, "message": "idle", "is_committed": true,
 "progress_percent": 0, "commit_timeout": 0, "partition": 0}
```

`state` is an int, the message key is `message`, progress is `progress_percent`, and there are three fields the docstring does not mention.

## What I would change

1. **`OTA.Start` → `OTA.Update`**, and take `file` and `ignore_same_version` while you are there, or leave them out deliberately. The parameters you already pass need no change.
2. **Correct the `OTA.Status` docstring** to the reply above.
3. **Wrap `OTA.Commit`** — this is the functional one rather than cosmetic. The PR says `commit_timeout` "exists for exactly that and is worth using", and it does, but with only `start` and `get_status` there is no way to commit: a caller who sets it gets a guaranteed rollback when the timer expires. `OTA.Revert`, `OTA.CreateSnapshot` and `OTA.GetBootState` are the natural companions — `GetBootState` answers unauthenticated and is a cheap read:

```
{"active_slot": 0, "is_committed": true, "revert_slot": 0, "commit_timeout": 0}
```

Happy to send this as a PR against your branch, same as #546 and #558 — say the word. And the caveat stands from your own PR body: I have verified names, signatures and idle replies, not an update in flight. A bad image is still how a grill stops working, and I have not been willing to try one.

dknowles2:
Rebased onto current `main` (`63f3fda`) at `1037352`. Ready for you to merge.

**The rebase itself was one import collision** — `main` had changed that line for #562's `RPCResult`/`as_dict`, this branch for `from .ota import OTA`. Both wanted; nothing else conflicted.

**One thing needed updating beyond the conflict.** `ota.py` was written when `send_command` returned `dict`, so both methods failed mypy against the current type:

```
ota.py:30: Incompatible return value type (got "dict | list | None", expected "dict")
ota.py:41: Incompatible return value type (got "dict | list | None", expected "dict")
```

Brought into line with the convention #551 and #562 settled on:

* `get_status()` uses `as_dict()` — it is a getter and callers want the object.
* `start()` returns `RPCResult`, deliberately. `OTA.*` is a Mongoose core service, so it is **not modelled in `fake_firmware/init.js`** and nobody has seen a real grill answer it. Promising a `dict` would be the same mistake as the original `send_command -> dict`: a shape asserted rather than observed. The docstring says so.

779 tests, ruff, `ruff format --check` and mypy clean.

## What is still true from before

The scope is unchanged from the narrowing: `ota.py` with `OTA.Start` and `OTA.Status`, wired as `PitBoss.ota`, and its tests. The nine loose RPC wrappers are gone — two of them (`Sys.Reboot`, `PB.WiFiAwakeWDT`) landed via #530, the latter renamed `request_fast_updates()` because the RPC's name describes something it does not do.

And the caveat I would not want lost if this merges: **nothing here demonstrates a real grill answering `OTA.Start`.** The wrappers match Mongoose's documented service and the tests pin the request shape, but that is the whole of the evidence. It is also the sharpest surface in the library — a bad image is how a grill stops working entirely, which is what `commit_timeout` exists for.

Since #561 landed, `list_rpcs()` is now the way a caller can check whether a given grill serves `OTA.*` at all rather than assuming it.

PR #491, opened 2026-08-01
Fix LBL/LFS parsing routines and mark those models supported

dknowles2:
## Summary

The Louisiana Grills boards (`LBL`, `LFS`) have specs in `grills.json`, but all nine of their models are filtered out by `UNSUPPORTED_MODELS` as "Bad parsing routine", so `get_grills(control_board="LBL")` returns nothing.

Context: [ha-pitboss#258](https://github.com/dknowles2/ha-pitboss/pull/258) works around this on the Home Assistant side by aliasing the `LBL-*` BLE name prefix to the `PBL` board. That gets past `unknown_grill`, but LBL and PBL don't share a frame layout, so it reads the wrong bytes.

Both routines were derived from the PitBoss ones without adjusting for the shorter LBL/LFS frame, which has no dedicated smoker temperature field. Two fields are dropped as a result.

## 1. `smokerActTemp` reads no bytes of its own

It duplicates another field's offset, and the vendor picked a different one per frame:

| | `FE0B` status | `FE0C` temperatures |
|---|---|---|
| `smokerActTemp` reads | offset 14 = **p4Temp** | offset 17 = **grillSetTemp** |

`PitBoss._on_state_received` merges both payloads into one state dict, so as shipped the value flips between probe 4's reading and the grill setpoint depending on which frame arrived last.

Neither reading survives a look at the probe counts in `grills.json`:

| board | models | `meat_probes` |
|---|---|---|
| LBL | LG0800BL, LG1000BL, LG1200BL, LG300BL, LGV4BL | 2 |
| LFS | LG800FL, LG800FP, LG1200FL, LG1200FP | 4 |

On LFS, `p4Temp` is a real meat probe, so reporting it as a chamber temperature would be wrong. On LBL, slots 3 and 4 are unused and read 960 → `None`, so it would report nothing. The field is dropped on both.

## 2. The status grill-temperature block overlaps the flag block

```js
moduleIsOn: parts[21] === 1,
err1:       parts[22] === 1,
...
switch (parts[23]) {
  case 1: status.grillSetTemp = convertTemperature(parts, 20); break;
  case 2: status.grillTemp    = convertTemperature(parts, 20); break;
}
```

`convertTemperature(parts, 20)` reads bytes 20, 21 and 22 — the same bytes `moduleIsOn` and `err1` claim. These frames are digit-per-byte, so that would constrain the grill temperature to 200/201/210/211 for `parts[20] == 2`. Both readings cannot be right, whatever the grill actually sends.

This is **dropped rather than rewritten to a guessed offset**. Grill temperatures come from the `FE0C` reply, at offsets 17 and 20, untouched by this PR — which is how every other board already works: `PBL`, `PBC` and `PBV` all ship this same `switch` commented out of their own status routines, and `ble.py` dispatches `FE0B` and `FE0C` as independent pushes.

## Where the fix lives

`grills.json` is regenerated wholesale from the vendor API by `scripts/update_grills.sh`, so a fix in the data file would be overwritten on the next sync. The removal happens in `ControlBoard.from_dict`, ahead of the existing `_scrub_js` pass:

```python
_BROKEN_BOARDS = ("LBL", "LFS")
_BROKEN_STATUS_FIELDS = ("smokerActTemp", "grillSetTemp", "grillTemp")
_BROKEN_TEMPERATURE_FIELDS = ("smokerActTemp",)
```

If Dansons ever fixes the routines upstream, this degrades to "no grill temperature in the status frame" — normal behavior for every other board — rather than silently shadowing their fix with a stale copy.

Separately, `_COMMAND_SLUG_OVERRIDES` normalizes the vendor's `set-prove-1-temperature` typo on the LBL board. Without it, `PitBoss.set_probe_temperature()` raises `KeyError` on these grills, and ha-pitboss silently drops the probe-1 target entity.

The nine `LG*` models come out of `UNSUPPORTED_MODELS`, leaving only the three genuinely nonstandard-format entries.

## Tests

`test_parse_state` asserts an exact dict. LBL and LFS are the only supported boards that report probe temperatures in the `FE0B` status frame — every other board has those fields commented out — so `want` now includes them when the routine provides them.

All 513 tests pass, with the nine LG models now exercised by the parametrized tests instead of skipped. `mypy`, `ruff check` and `ruff format --check` are clean.

```
LG0800BL probes=2   status: no smoker field   temps: grillSet=225 grill=215 p4=164
LG800FL  probes=4   status: no smoker field   temps: grillSet=225 grill=215 p4=164
```

## Note on the vendor data

I re-fetched `LG0800BL` and `LG800FL` from `api-prod.dansonscorp.com` while preparing this: `status_function`, `temperature_function` and the command list are all byte-identical to what's in the repo, so this isn't papering over an upstream fix.

Worth flagging separately: the nightly `update-grills.yml` job has been failing for months (it installs the now-deleted `requirements-dev.txt`, and before that died on missing `~/.pitboss` credentials), so `grills.json` hasn't been refreshed since 2025-12-02.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,15 @@ The following models should be supported. Note however that only the `PBV4PS2` m
 
 <!-- GRILLS START -->
 
+*  [LG0800BL](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG0800BL.png)
+*  [LG1000BL](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG1000BL.png)
+*  [LG1200BL](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG1200BL.png)
+*  [LG1200FL](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG1200FL.png)
+*  [LG1200FP](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG1200FP.png)
+*  [LG300BL](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG300BL.png)
+*  [LG800FL](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG800FL.png)
+*  [LG800FP](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LG800FP.png)
+*  [LGV4BL](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/LGV4BL.png)
 *  [Lexington Wi-Fi Upgrade](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/ghost%20grill.png)
 *  [PB0500SP](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/500sp-109.png)
 *  [PB0820SP/SPW](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/PB820sp.png)
@@ -170,6 +179,7 @@ The following models should be supported. Note however that only the `PBV4PS2` m
 *  [PB1000XL/PB1000SC3](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/PB1000SC3-Front-101619.png)
 *  [PB1000XLW1 (Austin XL)](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/austin-xl.png)
 *  [PB1020CS2](https://dansons-mobile.s3.dualstack.us-east-1.amazonaws.com/grill-images/PB1020CS2.1.png)
+*  [PB1020DX](Https://dansons-mobile.s3.dualstack.us-east-1.amazonaws.com/grill-images/PB1020DX.png)
 *  [PB1020NXW](https://dansons-mobile.s3.dualstack.us-east-1.amazonaws.com/grill-images/PB1020NX.png)
 *  [PB1100HTC](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/PB1100HTC.M-Line%20Heritage.png)
 *  [PB1100PS1](https://dansons-mobile.s3.us-east-1.amazonaws.com/grill-images/1100ps.png)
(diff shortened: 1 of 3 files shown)


dknowles2:
Added the regenerated `README.md` (`scripts/update_readme.py`) so the supported-model list matches `get_grills()`.

The generator already filters `UNSUPPORTED_MODELS` correctly, but the checked-in list had gone stale: 125 entries against 137 returned. The regeneration adds the 9 LG models this PR enables, plus three that were stranded by the 2025-12-02 `grills.json` update (`PB1020DX`, `PB1600PS4`, `PBV7P2`) and one corrected image URL for `PBV5 P2`. Happy to drop those four lines if you'd rather keep this PR to the LG models alone.

Root cause for the drift, separate from this PR: `update-readme.yml` fires only on `release: published`. It last succeeded on 2025-09-05, never ran for the 2025-12-29 `2025.12.0` release, and its 2026-07-23 run for `2026.7.0` failed with `fatal: could not read Username for 'https://github.com'` — an invalid or expired `secrets.PAT`. It also still installs `requirements.txt`, which the uv migration removed, so a future run would fail at the install step as well.

dknowles2:
Switched from rewriting the vendor JS to discarding the bad fields after parsing.

The JS munging was a leftover: the first draft rewrote byte offsets (`switch (parts[23])` -> `parts[20]`), which genuinely had to happen before dukpy evaluated the snippet. Once the approach became pure deletion, editing the JS stopped being necessary and only added risk — regex over vendor JS is brittle, and the CRLF line endings in `grills.json` had already made one pattern silently match nothing.

`parse_status` and `parse_temperatures` now filter the parsed dict against `DROPPED_STATUS_FIELDS` / `DROPPED_TEMPERATURE_FIELDS`. `_status_js_func` is once again byte-identical to what `grills.json` ships (verified), so diffing against a fresh API dump stays trivial and the override is data rather than transformed text.

One test consequence: the synthetic frame builder used to infer the layout from the JS text, which worked only because the munged JS no longer mentioned the dropped field. It now reads the same constants, which is more honest about what it's asserting — these frames reserve no bytes for those fields.

513 tests, mypy and ruff all pass.

PR #492, opened 2026-08-01
Repair the grills.json sync workflow

dknowles2:
## Summary

The nightly `update-grills.yml` job has failed **every run since 2025-10-16**, and `pytboss/grills.json` hasn't been refreshed since 2025-12-02 (#368, which was a manual local run). Five separate faults had stacked up.

> [!IMPORTANT]
> This needs two new repository secrets — `PITBOSS_USERNAME` and `PITBOSS_PASSWORD` — before the workflow can succeed. Without them it now fails with an explicit error instead of a traceback.

## What broke, in order

**Credentials (2025-10-16).** Last success was 2025-10-15; every run since failed. Those logs have aged out, but the repo's own history explains it: on 2025-12-01 you added an auth library (#359) and converted the scripts to async (#360), reintroducing the `~/.pitboss` credential file that `771b9bb` had removed in July. That's consistent with the Dansons API starting to require authentication around 2025-10-16.

From #360 onward the failure is `KeyError: 'pitboss'` — confirmed in runs on 2026-05-09 and 2026-07-19. The workflow has never had a step that writes `~/.pitboss`, and the repo's only secrets are `BOT_PAT`, `PAT` and `PYPI_TOKEN`.

**Dependencies (2026-07-24).** #484 migrated to uv dependency groups and deleted `requirements-dev.txt`, which this workflow still installs. Runs since then fail at the install step, before reaching the script at all.

**Change detection.** The step assigns `GRILLS_JSON` and then tests:

```bash
if ! git diff --exit-code "$FILE_TO_UPDATE"; then
```

`FILE_TO_UPDATE` is never set. An empty pathspec makes git exit 128 (`fatal: ambiguous argument ''`), and inside `if !` that makes `changes_detected` unconditionally `true`. Masked today because `create-pull-request` skips when there's no diff, but the guard does nothing.

**Atomicity.** `python -m scripts.dump_grills > "$GRILLS_JSON"` truncates the tracked file *before* the script runs, so any mid-run failure leaves it empty or half-written:

```
# old redirect, dump fails partway
after:  {"partial": tru
# new temp-file + mv
after:  {"real":"data"}
```

`scripts/update_grills.sh` had the same problem plus an exit-code bug:

```bash
python3 -m scripts.dump_grills >pytboss/grills.json
if [[ $? -ne 0 ]]; then exit $?; fi
```

The inner `$?` is the status of the `[[ ]]` test, not the script, so it exits **0** after truncating the file. That's the most likely origin of any zero-byte `grills.json` people have hit locally.

**Resilience.** The API hangs up on the occasional request during a full dump — I hit `ClientConnectionError: Connection closed` at `grill_id 67` while investigating — and nothing retries, so one hiccup aborts the run.

## Changes

- Write `~/.pitboss` from `PITBOSS_USERNAME` / `PITBOSS_PASSWORD`, with `umask 077`, an explicit error when unset, and an `always()` cleanup step.
- `uv sync --locked --group dev`, matching `build-and-test.yml`.
- `$FILE_TO_UPDATE` → `git diff --quiet -- "$GRILLS_JSON"`.
- Dump to a temp file, `mv` into place only on success — in both the workflow and `update_grills.sh`.
- Fix the exit-code bug in `update_grills.sh`.
- Retry dropped connections in `get_grill_details` with a short backoff, and re-raise non-404 response errors instead of falling through to `resp.json()` on an error body.
- Note in the generated PR body that new models sometimes ship parsing routines that fail the tests, so CI should be checked before merging.

## Verified

- Workflow YAML parses; step order and conditions check out (credentials are removed before the PR step, and on failure via `always()`).
- `bash -n` clean on `update_grills.sh`; `ruff`, `ruff format --check` and `mypy` clean.
- The run-script block was exercised locally against a git fixture: partial-dump failure leaves the tracked file untouched and exits non-zero, and change detection reports `false` when unchanged and `true` when changed.

Not verified end to end — that needs the secrets and a `workflow_dispatch` run.

## Two things worth deciding separately

`main` has no branch protection, so `gh pr merge --auto --squash` in `auto-merge.yml` isn't actually gated on the test run for `dknowles-bot` PRs. Given that new models can fail the suite, requiring the `build` / `mypy` / `ruff` checks on `main` would make the auto-merge path safe.

`update-readme.yml` is broken too, separately: it installs the also-deleted `requirements.txt`, and its last run (2026-07-23, release `2026.7.0`) failed with `fatal: could not read Username for 'https://github.com'` — an invalid or expired `secrets.PAT`. Left alone here to keep this PR to one workflow.


dknowles2:
Added a second commit dropping the schedule from daily to weekly — Sundays at 00:00 UTC. `workflow_dispatch` is untouched, so on-demand runs still work.

The catalogue changes rarely (last real movement was 2025-12-02), so a nightly sweep of 149 sequential API requests wasn't buying much.

One thing I left alone: the runs consistently start 50–110 minutes late (scheduled 00:00, observed 00:50–01:55), which is the usual congestion at the top of the hour on shared runners. An offset like `17 4 * * 0` would land closer to on time if that matters, but it doesn't affect correctness.

dknowles2:
Moved the weekly run to Thursdays (`0 0 * * 4`) so there's a couple of working days plus the weekend to deal with anything a refresh breaks.

Note on the local timing: 00:00 UTC Thursday is Wednesday evening in US timezones (~20:00 EDT), and the runs historically fire 50–110 minutes late, so in practice the PR will be waiting first thing Thursday morning. If you'd rather it land during Thursday itself, `0 12 * * 4` would put it around 08:00 EDT.

dknowles2:
Correction to this PR's description: the claim that `main` has no branch protection is wrong.

I checked `repos/dknowles2/pytboss/branches/main/protection`, which returns 404 when protection comes from a *ruleset* rather than classic branch settings. There is an active ruleset targeting the default branch:

```
pull_request              required_approving_review_count: 0
required_status_checks    build, mypy, ruff
deletion, non_fast_forward, copilot_code_review
bypass_actors: []
```

So `gh pr merge --auto --squash` in `auto-merge.yml` **is** gated on the test run — required checks must pass before it merges, and `required_approving_review_count: 0` means bot pull requests do not stall waiting for a review. No change is needed there.

The same ruleset does affect `update-readme.yml`, which pushes directly to `main`; that is handled in #498.

PR #497, opened 2026-08-01
Preserve control board variants when two rows share a model name

### LINKED ISSUE #496: dump_grills silently discards control board variants when two vendor rows share a model name
## Summary

`scripts/dump_grills.py` collects rows into a dict keyed by model name:

```python
grills[grill["name"]] = grill
```

The vendor holds **147 rows for 140 model names**. Seven models exist twice, on two different control board generations, and the higher grill ID silently wins. The other board's definition is discarded.

## What gets discarded

| model | discarded board | kept board |
|---|---|---|
| PB1600PS2 | PBL | PBP |
| PB1500NXW, PB850M, PB850CS2 | PBM | PBM2 |
| PB850DX, PB1150DX | **PBL2** | PBL3 |
| PBV3NX | PBV | PBV2 |

`PBL2` appears in no other row, so it is **absent from `grills.json` entirely** — 19 boards are present out of the 20 the API serves.

## Consequence 1: models are filed under a board they may not advertise as

ha-pitboss derives the control board from the BLE name prefix and calls `get_grills(control_board=...)` to populate the model dropdown. Because these seven models are filed under the newer board, they are missing from the older board's list:

- a device advertising `PBM-…` will not find `PB850M`, `PB850CS2` or `PB1500NXW`
- a device advertising `PBV-…` will not find `PBV3NX`
- a device advertising `PBL-…` will not find `PB1600PS2`
- a device advertising `PBL2-…` finds nothing at all, since the board does not exist

That last case is why [ha-pitboss](https://github.com/dknowles2/ha-pitboss) carries a `PBL2 -> PBL3` alias in its config flow.

## Consequence 2: the PBL2 alias misparses in celsius

Diffing the discarded boards against the kept ones, three of the four differ only in comment formatting — `//` versus `/* */`, whitespace, and where a block comment opens. Functionally identical.

`PBL2` versus `PBL3` is not:

```diff
-console.log("Temp function", temps, temps.isFahrenheit);
+/*
 if (!temps.isFahrenheit) {
   ...
-  console.log("Temp function - converted", temps);
 }
+*/
```

`PBL3` runs the fahrenheit-to-celsius conversion in its `temperature_function`; `PBL2` has it commented out, meaning that board reports the correct unit already. A PBL2 device parsed with PBL3's routine therefore gets a second conversion applied whenever the grill is set to celsius — 225 reported as 107.

In fahrenheit the block is skipped and behaviour is identical, which is presumably why this has gone unnoticed.

`PBV` versus `PBV2` also differs substantively in its command list — `PBV2` lacks `turn-light-on`, `turn-light-off`, `turn-primer-motor-on` and `turn-primer-motor-off`. No live path today: all five PBV2 models have `lights=0`, so `PitBoss.turn_light_on` returns early, and ha-pitboss gates the primer switch on command presence. It would raise `KeyError` if a PBV2 model ever shipped with lights.

## Suggested fix

Preserve the shadowed rows rather than overwriting them, and let callers disambiguate:

- keep both rows in `grills.json` so `get_grills(control_board=...)` sees every board a model ships on
- keep `get_grills()` unfiltered returning one entry per name, so the README and the model dropdown are unchanged
- give `get_grill()` an optional `control_board` argument, so a device that advertises `PBL2-…` resolves to the PBL2 definition rather than the PBL3 one

That makes the `PBL2 -> PBL3` alias in ha-pitboss unnecessary, and stops a definitions refresh from quietly deciding which of two board generations a model belongs to.

## Note

`control_board.device_status_function` is a new key in the vendor data as of this year. It is null on all 147 rows, including the shadowed ones, so it is not currently useful for anything.

dknowles2:
Fixes #496.

## Problem

`dump_grills` keyed collected rows by model name alone:

```python
grills[grill["name"]] = grill
```

The vendor serves **147 rows for 140 names**. Seven models exist on two control board generations, so the higher grill ID silently overwrote the other board's definition. `PBL2` appears in no other row, so it was missing from `grills.json` entirely — 19 boards present out of the 20 the API serves.

| model | discarded | kept |
|---|---|---|
| PB1600PS2 | PBL | PBP |
| PB1500NXW, PB850M, PB850CS2 | PBM | PBM2 |
| PB850DX, PB1150DX | **PBL2** | PBL3 |
| PBV3NX | PBV | PBV2 |

**Models were filed under a board they may not advertise as.** ha-pitboss derives the board from the BLE name prefix, so a grill announcing `PBM-…` would not find `PB850M` in the model list, and one announcing `PBL2-…` found nothing at all — hence the `PBL2 -> PBL3` alias in its config flow.

**That alias misparses.** Three of the four discarded boards differ from their replacement only in comment formatting. `PBL2` versus `PBL3` does not:

```diff
-console.log("Temp function", temps, temps.isFahrenheit);
+/*
 if (!temps.isFahrenheit) {
   ...
 }
+*/
```

`PBL3` runs the fahrenheit-to-celsius conversion; `PBL2` has it commented out because that board already reports the right unit. A PBL2 grill set to celsius therefore gets converted twice — 225 shown as 107. In fahrenheit the block is skipped, which is why this went unnoticed.

## Change

Keep both rows: the higher ID under the plain model name, the other under `"<name> (<board>)"`.

- `get_grills()` unfiltered yields each model once, so callers listing supported models see no duplicates — **the README regenerates with no diff**.
- `get_grills(control_board=...)` now sees every board a model ships on. `PBL2` returns `PB850DX` and `PB1150DX`; `PBM` gains `PB850M`, `PB850CS2`, `PB1500NXW`; `PBV` gains `PBV3NX`; `PBL` gains `PB1600PS2`.
- `get_grill()` takes an optional `control_board` to select between variants. Without it, behaviour is unchanged — `get_grill("PB850DX")` still returns the PBL3 variant.

Once this lands, ha-pitboss can drop the `PBL2 -> PBL3` alias and pass the advertised prefix to `get_grill()` instead.

## Data change is additions only

```
pre-existing entries modified: none
entries removed: none
entries added  : ['PB1150DX (PBL2)', 'PB1500NXW (PBM)', 'PB1600PS2 (PBL)',
                  'PB850CS2 (PBM)', 'PB850DX (PBL2)', 'PB850M (PBM)', 'PBV3NX (PBV)']
totals: 140 -> 147
```

The line-level diff shows deletions, but those are alignment artifacts — `"PB850DX (PBL2)"` sorts adjacent to `"PB850DX"`. Verified semantically above.

The vendor has also accumulated eight months of metadata changes. They're inert — no parsing routine or command list differs on any of the 140 models — and belong in the refresh the sync workflow will produce, not in this diff.

## Two test-harness predicates the new variants exposed

Both read commented-out code as live, and nothing caught it because PBL2 did not exist in the data until now.

`JSFunc.has_key` rejected any field named inside a comment. PBL2 comments out a conversion block that mentions `p4Temp` while parsing it live, so the frame builder omitted those bytes and every later offset shifted. Plain membership in comment-stripped source is wrong too: `LFS` comments `p1Target` out of its object literal but keeps a live `status.p1Target = ftoc(status.p1Target)`, which converts `undefined` and reports nothing. The predicate now asks whether live code *assigns* the field in the returned object.

`converts_to_celsius` (added in #495) had the mirror problem — it matched PBL2's commented-out `ftoc` block and expected a conversion that never runs. That block being commented out is precisely what distinguishes PBL2 from PBL3.

Both now share a helper that strips comments first.

Parametrized tests cover every model/board pairing rather than one entry per model — which is what left `PBL2` untested through all of this.

## Verified

534 tests pass, up from 513: exactly the 7 new variants × 3 parametrized tests. `ruff`, `ruff format --check` and `mypy` clean. `scripts/update_readme.py` regenerates with no diff.


dknowles2:
Retargeted from `tolerate-grill-api-errors` to `main`: `build-and-test.yml` only runs on pull requests targeting `main`, so as a stacked PR this got no CI at all.

The consequence is that #494's commit now appears in this PR's diff as well. The two are still logically separate, and this one genuinely depends on #494 — without its 500 tolerance a full sweep can't complete, which is what regenerating `grills.json` required.

Merge #494 first and this diff collapses to just its own commit. If you'd rather merge this one first, #494 comes along with it and can be closed.

dknowles2:
Rebased onto `main` now that #493, #494 and #495 have landed. This PR is back to a single commit — the duplicate of #494's work is gone.

The rebase surfaced one more thing. #495's `converts_to_celsius` tests `"ftoc" in self._js` against the raw source, and PBL2 ships the same conversion block as PBL3 with the whole thing commented out. So the predicate matched a block that never runs and the test expected a conversion PBL2 doesn't perform — two failures. That block being commented out *is* the difference between the two boards, so it isn't an edge case, just one that nothing exercised until PBL2 existed.

Both predicates now share a helper that strips comments first. 534 tests pass, `ruff`, `ruff format --check` and `mypy` clean.

PR #501, opened 2026-08-01
Drop vendor bookkeeping fields from the checked-in definitions

dknowles2:
## Summary

`created_at`, `updated_at` and `deleted_at` sit on every grill, every control board and every one of their commands. `app_layout` is `"standard"` on all 147 rows. `manual_url` is a product support link. Nothing in pytboss, its scripts or ha-pitboss reads any of them.

Dropping them at dump time:

```
before:  336,948 bytes
after:   268,213 bytes   (20% smaller)
```

on top of the 72% #500 saved by storing boards once. Together the definitions have gone from 1,185,170 bytes to 268,213 — a **77% reduction**.

The timestamps are the bulk of it, mostly because they repeat on all ~200 command objects; grill and board timestamps alone would only be 14%.

## Verified

- **Nothing consumes the dropped fields** — grepped `pytboss/`, `scripts/`, `tests/` and ha-pitboss's `custom_components/pitboss/`.
- **Everything retained is untouched**: grill and board counts unchanged, and no surviving value differs from before. None of the dropped keys remains anywhere, at any of the three levels.
- **A fresh `python -m scripts.dump_grills` against the live API reproduces the committed file exactly**, so the script genuinely emits this shape rather than the transformation having been applied by hand.
- 534 tests pass; `ruff`, `ruff format --check`, `mypy` clean; `scripts/update_readme.py` regenerates `README.md` with no diff.

## What this does not do

It does not meaningfully quieten the automated refreshes. I checked against the #499 data: the vendor also materialised `mpc_type` and `device_status_function` as null columns in that refresh, so one still touches most rows.

```
entries differing in the #499 refresh
  untrimmed:                     140 of 147
  with this change:              140 of 147
  + mpc_type, device_status_fn:    6 of 147
```

Both of those are `null` on every row and referenced nowhere — not in pytboss, not in ha-pitboss, and not in any vendor firmware build I checked (0.2.3, 0.5.7, 0.6.0). Adding them would cost another 4 KB and collapse a refresh to 6 entries, but that is a separate call and this PR deliberately sticks to the fields agreed.

### DIFF
--- a/pytboss/grills.json
+++ b/pytboss/grills.json
@@ -3,2609 +3,1774 @@
     "LBL": {
       "control_board_commands": [
         {
-          "control_board_id": 8,
-          "created_at": "2022-12-06T15:28:54.000000Z",
-          "deleted_at": null,
           "description": null,
           "function": null,
           "hexadecimal": "FE0902FF",
           "id": 95,
           "name": "Set Temperature to Celcius",
-          "slug": "set-celsius",
-          "updated_at": "2022-12-09T12:11:06.000000Z"
+          "slug": "set-celsius"
         },
         {
-          "control_board_id": 8,
-          "created_at": "2022-12-06T15:31:52.000000Z",
-          "deleted_at": null,
           "description": null,
           "function": null,
           "hexadecimal": "FE0901FF",
           "id": 96,
           "name": "Set Temperature to Fahrenheit",
-          "slug": "set-fahrenheit",
-          "updated_at": "2022-12-06T15:31:52.000000Z"
+          "slug": "set-fahrenheit"
         },
         {
-          "control_board_id": 8,
-          "created_at": "2022-12-06T15:33:02.000000Z",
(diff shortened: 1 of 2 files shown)


dknowles2:
Added `control_board_id` to the command-level drops — commands are stored inside the board they belong to, so naming it again on each one adds nothing their position doesn't.

```
336,948  before this PR
268,213  timestamps, app_layout, manual_url
261,729  + command.control_board_id      (-6,484)
```

22% for the PR, and 1,185,170 → 261,729 combined with #500 — **78% overall**.

Re-verified after the change: a fresh dump against the live API still reproduces the committed file exactly, 534 tests pass, ruff/mypy clean, README regenerates with no diff.

One adjacent case I did not touch: grills carry a top-level `control_board_id` too, which is equally redundant now that they reference the board by name — worth another 4,299 bytes. Say the word and I'll add it.

PR #502, opened 2026-08-01
Say which models and boards a definitions refresh changes

dknowles2:
## Summary

The sync workflow opened every pull request with the same fixed body, so reviewing one meant reading a diff that can run to hundreds of lines just to learn whether a model appeared or disappeared. That summary now goes in the description:

```markdown
Update grills.json with the latest definitions.

### Adds grill models

* PB850DX (PBL2)

### Removes grill models

* PBV3M

### Adds control boards

* PBL2

Newly added models sometimes ship parsing routines that fail the
test suite, so check CI before merging rather than assuming a
definitions-only change is safe. A refresh with no headings above
changed only existing definitions.
```

## One implementation, not two

`scripts/update_grills.sh` already computed exactly this for the commit message of a local update. Rather than reimplement it in YAML — where it would quietly drift from the shell version — it moves to `scripts/grills_diff.py` and both callers use it.

The shared version also reports **control boards**, which became worth naming separately once #500 stored them once instead of inlining them per model. A new board is arguably the more significant event: it is what ha-pitboss matches the BLE advertisement prefix against, and a missing one is what #497 was about.

Files predating that split are a flat map of models, so `_names` treats one as its own grills section rather than reporting all 140 models as removed.

## Verified

Exercised against real before/after data rather than fixtures:

| case | result |
|---|---|
| pre-#497 vs current | correctly lists the 7 added variants, plus all 20 boards (the one-off format transition) |
| the actual #499 refresh | empty — it added and removed nothing |
| identical files | empty |
| wrong argument count | usage message, exit non-zero |

`get_diff` in `update_grills.sh` was tested end to end in a throwaway repo against a staged change, since CI never runs that path — it reported an added model, a removed model and an added board correctly.

`bash -n` clean, workflow YAML parses with the interpolation in place, 534 tests pass, `ruff`, `ruff format --check` and `mypy` clean.

## Note

The summary is captured *before* the new definitions are moved into place, so it compares the committed file against the fresh dump. If a refresh changes only existing definitions the interpolation is empty, which the surrounding wording accounts for.


dknowles2:
Fixed a formatting bug I only caught by rendering the composed body rather than just the summariser's output.

The last bullet ran straight into the following paragraph:

```
* PBL2
Newly added models sometimes ship...
```

Markdown lazy continuation would have pulled that paragraph inside the list item. `summarize()` now returns text with no surrounding blank lines and the template supplies the separation, so the caller decides how it sits in a body or a commit message. Command substitution strips trailing newlines anyway, so relying on one from the script was never going to work.

Rendered both cases end to end after the fix — additions are correctly separated, and with nothing added or removed the script emits **0 bytes**, leaving blank lines that Markdown collapses.

Issue #509, opened 2026-08-02
set_grill_temperature clamps celsius values against fahrenheit bounds


trailro:
```python
# TODO: Clamp to a value from self.spec.temp_increments.
if self.spec.max_temp:
    temp = min(temp, self.spec.max_temp)
if self.spec.min_temp:
    temp = max(temp, self.spec.min_temp)
```

`spec.min_temp` / `max_temp` are always Fahrenheit. On a grill running in Celsius every setpoint below 180 is clamped **up** to 180 — asking for 105 °C lands on 190 °C on my PB1600PS1, and nothing lower can be set at all. Verified against hand-built MCU commands, which land exactly.

Snapping to `spec.temp_increments` instead, as the TODO suggests, also subsumes the clamping: a value past either end of the list lands on the nearest end.

Two details worth folding in:

- The unit has to come from the grill's own `isFahrenheit`, which means this depends on the cache actually being populated — see the `get_state()` issue.
- Deriving the Celsius list should **floor**, not round: the boards that convert inside their own parsing routine use `Math.floor((temp - 32) / 1.8)`, so to the grill 190 °F is 87 °C, not 88. Rounding advertises nine values the panel never lands on.

Branch with tests: [`trailro/pytboss:fix/set-grill-temperature-honours-the-grills-unit`](https://github.com/trailro/pytboss/tree/fix/set-grill-temperature-honours-the-grills-unit). It also adds `Grill.celsius_temp_increments` (two models publish one) and a public `accepted_setpoints()`. Split out of #506.

Issue #514, opened 2026-08-02
No public wrapper for set-celsius / set-fahrenheit


trailro:
Every board declares `set-celsius` and `set-fahrenheit`, but neither is reachable without `_send_command`. The grill's own unit — the one on its panel, and the one the values it reports are in — can be observed through `isFahrenheit` but not changed.

This is the last thing blocking a temperature-unit entity in ha-pitboss, since that integration uses no private pytboss APIs anywhere and I would rather not be the first.

Branch with a test: [`trailro/pytboss:feat/set-temperature-unit`](https://github.com/trailro/pytboss/tree/feat/set-temperature-unit) — `set_temperature_unit(fahrenheit: bool)`, same shape as the other command wrappers. Split out of #506.

PR #516, opened 2026-08-02
Update the cached state from get_state()

### LINKED ISSUE #508: get_state() does not update the cached state
`get_state()` parses the reply but never folds it into `self._state`, which is only ever written by pushed updates. On a connection that polls rather than subscribes, the cache therefore stays empty for the whole session, and anything reading it sees nothing.

That matters beyond the cache itself: it is where `isFahrenheit` lives, so the library has no way to know what unit the grill is working in unless pushes happen to be arriving.

The reply already carries exactly the fields the push path parses, so folding it in is a couple of lines. `get_state()` would still return a live read rather than the cache; only the side effect is new.

Branch with a test: [`trailro/pytboss:fix/get-state-updates-cached-state`](https://github.com/trailro/pytboss/tree/fix/get-state-updates-cached-state). Split out of #506.

trailro:
Fixes #508.

`self._state` is only ever written by `_on_state_received`, so on a connection that polls rather than subscribes it stays empty for the whole session. `get_state()` parses exactly the same fields and throws them away afterwards.

The immediate consequence is that `isFahrenheit` is unknowable on such a connection, which is what #509 needs to send a setpoint in the grill's own unit — but the cache being silently empty seems worth fixing on its own regardless.

`get_state()` still returns a live read rather than the cache; only the side effect is new, and the docstring says so. Guarded on a non-empty parse so a malformed reply cannot wipe what is already there.

One test, which fails on `main`: after `get_state()`, `_state` matches the parsed reply.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -246,15 +246,19 @@ class PitBoss:
     async def get_state(self) -> StateDict:
         """Issues a live RPC to fetch and return the current grill state.
 
-        Unlike `subscribe_state()`, this does not use the cached state
-        maintained from prior pushed updates; it always queries the grill
-        directly.
+        Unlike `subscribe_state()`, this always queries the grill rather than
+        returning the cached state. The reply does update that cache, so
+        anything relying on it stays correct on a connection that only ever
+        polls.
         """
         resp = await self._conn.send_command(
             "PB.GetState", await self._authenticate({})
         )
         status = self.spec.control_board.parse_status(resp["sc_11"]) or {}
         status.update(self.spec.control_board.parse_temperatures(resp["sc_12"]) or {})
+        if status:
+            async with self._lock:
+                self._state.update(status)
         return status
 
     async def get_firmware_version(self) -> dict:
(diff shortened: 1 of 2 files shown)


trailro:
`AGENTS.md` asks for a label so release-drafter can categorise it — I tried to attach `bug` but do not have permission to label in this repo, so that one is on you.

Happy to open the rest of the split-out issues as PRs at whatever pace suits. #514 is the smallest and #509 is the one that depends on this landing first; the others are independent of each other.

dknowles2 (commented):
*(Automated review via Hermes Reviewer Agent)*

Reviewed the diff and ran the tests locally (`uv run pytest`, all 52 pass, including the new `test_get_state_updates_the_cached_state`).

The fix is small, correct, and follows the existing convention in this file: it takes `self._lock` and calls `self._state.update(status)` the same way `_on_state_received` already does, guarded on a non-empty parse so a malformed reply can't wipe existing cached state. Docstring is updated to match the new behavior. Good regression test that reproduces the bug on `main` (`_state` stays `{}` after `get_state()`) and verifies the fix.

No concerns. LGTM.

Recommendation: Approve & Merge

PR #517, opened 2026-08-02
Snap the grill setpoint to an accepted value in the grill's own unit

### LINKED ISSUE #509: set_grill_temperature clamps celsius values against fahrenheit bounds

trailro:
Fixes #509.

`set_grill_temperature` clamped against `min_temp`/`max_temp`, which are Fahrenheit, so a Celsius grill had a floor of 180 and a ceiling of 500 in a range that tops out around 260. The `TODO` above it already said the right answer was `temp_increments`.

This replaces the clamp with a snap to the nearest accepted setpoint, in whichever unit the grill is currently working in, and exposes `accepted_setpoints(fahrenheit=...)` so a caller can present the same list.

Two details worth flagging, both verified against real hardware (PBV4PS2, Pit Boss 1600 Pro Elite):

**The Celsius list is truncated, not rounded.** The boards that convert internally do `Math.floor((temp - 32) / 1.8)`, so 190F is 87C to the grill, not 88. Rounding gets 9 of my grill's 19 Celsius setpoints wrong — I found this by sending 88 and reading back 87. Where `grills.json` declares its own `celsius_temp_increment`, that wins over the derivation; `Grill` didn't parse that field before, so this adds it.

**Unit inference.** I picked `_state["isFahrenheit"]` over adding an explicit argument, because every existing caller passes a temperature in the unit the grill is displaying and an argument would change that contract. This is the choice I asked about in the issue and haven't heard back on — if you'd rather have `set_grill_temperature(temp, fahrenheit=True)`, say so and I'll reshape it; the snapping logic is unaffected either way.

The one soft spot: before the first poll `_state` is empty and it assumes Fahrenheit. That is no worse than today, where the Fahrenheit bounds were applied unconditionally, but it is an assumption rather than a fact. #516 landing means a polling-only transport now populates `_state` on the first `get_state()`, which is why this one was waiting on it.

`pytest` (692), `ruff` and `mypy` clean.

`AGENTS.md` asks for a label; I still can't set one here, so `bug` is on you.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -5,6 +5,7 @@ import inspect
 import json
 import logging
 from collections.abc import Awaitable, Callable
+from math import floor
 from time import monotonic
 
 from .codec import encode, timed_key
@@ -198,16 +199,33 @@ class PitBoss:
         )
         self._password = new_password_bytes
 
+    def accepted_setpoints(self, fahrenheit: bool = True) -> list[int]:
+        """Grill setpoints the control board honours.
+
+        The board ignores anything that is not on this list.
+
+        :param fahrenheit: Whether to return the list in Fahrenheit.
+        """
+        increments = self.spec.temp_increments or []
+        if fahrenheit:
+            return list(increments)
+        if celsius := self.spec.celsius_temp_increments:
+            return list(celsius)
+        # Match the boards that convert inside their own parsing routine:
+        # they floor, so 190F is 87C to the grill rather than 88.
+        return [floor((v - 32) / 1.8) for v in increments]
+
     async def set_grill_temperature(self, temp: int) -> dict:
         """Sets the target grill temperature.
 
-        :param temp: Target grill temperature.
+        Snapped to the nearest value the control board accepts, expressed in
+        the unit the grill is currently working in.
+
+        :param temp: Target grill temperature, in the grill's own unit.
         """
-        # TODO: Clamp to a value from self.spec.temp_increments.
-        if self.spec.max_temp:
-            temp = min(temp, self.spec.max_temp)
-        if self.spec.min_temp:
-            temp = max(temp, self.spec.min_temp)
+        fahrenheit = self._state.get("isFahrenheit", True)
+        if accepted := self.accepted_setpoints(fahrenheit):
+            temp = min(accepted, key=lambda value: abs(value - temp))
         return await self._send_command("set-temperature", temp)
 
     async def set_probe_temperature(self, temp: int) -> dict:
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -389,7 +389,11 @@ class Grill:
     """The number of meat probes available on the grill."""
 
     temp_increments: list[int] | None = field(default_factory=list)
-    """Supported temperature increments."""
+    """Supported temperature increments, in Fahrenheit."""
+
+    celsius_temp_increments: list[int] | None = field(default_factory=list)
+    """Supported temperature increments in Celsius, where the grill declares
+    its own list. Most do not; theirs is derived from `temp_increments`."""
 
     json: dict[str, Any] = field(default_factory=dict)
     """The raw JSON returned by the PitBoss API."""
@@ -400,7 +404,8 @@ class Grill:
 
         :param grill_dict: A top-level entry from `grills.json`, with `name`,
             `control_board`, `lights`, `has_mpc`, `meat_probes`,
-            `temp_increment`, `min_temp`, and `max_temp` keys.
+            `temp_increment`, `celsius_temp_increment`, `min_temp`, and
+            `max_temp` keys.
         """
         min_temp = None
         try:
@@ -424,6 +429,11 @@ class Grill:
             max_temp=max_temp,
             meat_probes=grill_dict["meat_probes"],
             temp_increments=[int(t) for t in grill_dict["temp_increment"].split("/")],
+            celsius_temp_increments=[
+                int(t)
+                for t in (grill_dict.get("celsius_temp_increment") or "").split("/")
+                if t.strip().isdigit()
+            ],
             json=grill_dict,
             control_board=ControlBoard.from_dict(grill_dict["control_board"]),
         )
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -391,12 +391,16 @@ class Grill:
     temp_increments: list[int] | None = field(default_factory=list)
     """Supported temperature increments, in Fahrenheit."""
 
-    celsius_temp_increments: list[int] | None = field(default_factory=list)
-    """Supported temperature increments in Celsius, where the grill declares
-    its own list. Most do not; theirs is derived from `temp_increments`."""
-
     json: dict[str, Any] = field(default_factory=dict)
     """The raw JSON returned by the PitBoss API."""
+
+    celsius_temp_increments: list[int] | None = field(default_factory=list)
+    """Supported temperature increments in Celsius, where the grill declares
+    its own list. Most do not; theirs is derived from `temp_increments`.
+
+    Declared after `json` rather than beside `temp_increments`: `Grill` is not
+    `kw_only`, so inserting a field earlier would shift every positional
+    argument after it and silently change what an existing caller passes."""
 
     @classmethod
     def from_dict(cls, grill_dict) -> "Grill":
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -187,6 +187,8 @@ def my_grill(
         min_temp=kwargs.get("min_temp", None),
         max_temp=kwargs.get("max_temp", None),
         has_lights=kwargs.get("has_lights", False),
+        temp_increments=kwargs.get("temp_increments", []),
+        celsius_temp_increments=kwargs.get("celsius_temp_increments", []),
     )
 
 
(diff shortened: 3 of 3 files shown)


dknowles2 (commented):
Not blocking this — the logic is right and the bug it fixes is real. But I found something while checking it that you could not have known, and it means one branch of this PR is currently dead code. That is my doing, not yours.

**`celsius_temp_increment` no longer exists in `grills.json`. I removed it in #504.**

So `grill_dict.get("celsius_temp_increment")` is always `None`, `celsius_temp_increments` is always `[]`, and `accepted_setpoints(fahrenheit=False)` always falls through to the derived list. `test_set_grill_temperature_prefers_a_declared_celsius_list` passes only because the fixture injects the value directly; nothing in the shipped catalogue can reach that branch.

Worse, the comment I wrote justifying the removal is simply wrong. It says the field *"only restates `temp_increment` in celsius, which TemperatureConverter already derives"*. It does not. The two grills that declare one, `PBV30DS` and `PBV30DX`, declare:

```
40, 45, 50, ... 150      (5 degree steps)
```

while the derivation produces:

```
37, 40, 43, 46, 48, 51, ... 148   (irregular)
```

Not a restatement — a different list, with a different step and different endpoints. I checked all 147 models rather than the two: those are the only ones affected, but for them the derived list is wrong at nearly every entry, which is presumably why the vendor bothered to publish one.

This is also live in ha-pitboss today. `coordinator.accepted_setpoints` reads `spec.json["celsius_temp_increment"]` for exactly this case, and since 2026.8.2 that key has been absent, so those two grills are already being offered setpoints their board will ignore. Your PR did not introduce it; it inherited it.

I will fix the data side — restore the field to `_DROPPED_GRILL_FIELDS`, correct that comment, regenerate. Once that lands, your `celsius_temp_increments` branch starts working and the test stops being hypothetical. Nothing for you to change here unless you want to rebase onto it.

**What I verified in the change itself**

* The floor-versus-round point is the good part of this PR and matches what the boards actually do — `floor((F - 32) / 1.8)` is the same conversion `ha-pitboss` derives its list with, and finding it by sending 88 and reading back 87 is better evidence than reading the JS.
* Replacing the `min_temp`/`max_temp` clamp with a snap is a strict improvement, and the two rewritten tests still cover above-range and below-range. One consequence worth stating: a grill with an empty `temp_increments` now gets no clamping at all, where it previously got the Fahrenheit bounds. Nothing in the catalogue has an empty list, so this is theoretical.
* The Celsius test is honest about what it asserts — `180/190/200F` flooring to `82/87/93C`, then snapping 88 to 87 rather than 88, which is exactly the bug.

**The question you asked, and did not get an answer to**

On `_state["isFahrenheit"]` versus an explicit `set_grill_temperature(temp, fahrenheit=True)`: that is David's call on the library's API shape, not mine, and I do not want to answer it for him by approving. I will flag it for him. For what it is worth I think your reasoning is sound — every existing caller passes a temperature in the unit the grill is displaying, so an argument would silently change what those callers mean, which is the worse failure.

The pre-first-poll assumption is genuinely no worse than the status quo, and #516 shrinks the window further. Fine as is.

Holding at comment rather than approve only because of the open API question. The code I would take.

trailro:
Understood — I'll leave the branch as is and rebase once the data fix lands, so the test stops being hypothetical.

Verified the parts I could rather than take them: `celsius_temp_increment` is present on 0 of the 147 entries in the shipped 2026.8.2 catalogue, and `ha-pitboss` does read it, at [`coordinator.py:55`](https://github.com/dknowles2/ha-pitboss/blob/main/custom_components/pitboss/coordinator.py) — `spec.json.get("celsius_temp_increment") or ""` — falling through to the derived list exactly as you describe. So PBV30DS and PBV30DX are being offered setpoints their boards ignore today, in the released integration, and I have that same fallback running on my own install.

That is a better find than the bug this PR fixes. Mine is wrong bounds; that one is wrong values on a specific pair of grills, silently, with a comment in the repo asserting it cannot happen. Happy to send the `ha-pitboss` side once `celsius_temp_increments` is real — it becomes `spec.celsius_temp_increments` there and the raw-JSON read goes away, which is the version-safety problem in reverse.

On the empty-`temp_increments` case: agreed it is theoretical, and I would rather it stayed a fall-through than grow a guard for a shape nothing in the catalogue has.

And thank you for not answering the `isFahrenheit` question on David's behalf — that is the right call, and it is the same line you drew on #333. I would rather wait than have it settled by an approval that did not mean to settle it.

dknowles2:
Labelled `bug` — the Fahrenheit bounds on a Celsius grill are a defect, not a shortfall. You can stop apologising for not being able to set them; that is on us, and I have gone through the rest of your open PRs too.

While deciding whether this also needed `breaking`, I found something I should have caught in the review.

**`celsius_temp_increments` is inserted before `json`, which reorders the `Grill` constructor.**

```python
@dataclass(frozen=True)   # not kw_only
class Grill:
    ...
    temp_increments: list[int] | None = field(default_factory=list)
    celsius_temp_increments: list[int] | None = field(default_factory=list)   # new, position 9
    json: dict[str, Any] = field(default_factory=dict)                        # was 9, now 10
```

`Grill` is not `kw_only`, so positional construction is legal, and anyone doing `Grill(name, board, ..., temp_increments, raw)` positionally now silently passes their raw dict as `celsius_temp_increments` and gets an empty `json`. No error, no type complaint at runtime — just a `Grill` whose `json` is `{}`.

Nothing in either repo does this: `from_dict` is the real constructor and `tests/test_api.py:182` uses keywords. But [REVIEW.md](https://github.com/dknowles2/pytboss/blob/main/REVIEW.md#releasing) names this case exactly — *"the dataclasses in `grills.py` are public, so removing an attribute or reordering a constructor is a breaking change even when nothing in either repo used it"* — and it is the kind of break that shows up as a bug report about missing `json` data months later rather than as a traceback.

Every field involved has a default, so the ordering is free. Moving it to the end costs nothing and keeps this a non-breaking change:

```python
    temp_increments: list[int] | None = field(default_factory=list)
    json: dict[str, Any] = field(default_factory=dict)
    celsius_temp_increments: list[int] | None = field(default_factory=list)
```

If you would rather keep it next to `temp_increments`, where it admittedly reads better, that is defensible — but then this wants the `breaking` label as well, and I would rather ship the reorder than the label.

Roll it into the rebase you are already planning for the `celsius_temp_increment` data fix; no need for a push just for this.

dknowles2:
Both answers from David, and the data fix is up.

**Keep the `isFahrenheit` inference.** Your reasoning carried it: every existing caller passes a temperature in the unit the grill is displaying, so an argument would silently change what those calls mean rather than clarify them. No signature change. The pre-first-poll assumption stays as is — it is no worse than the unconditional fahrenheit bounds it replaces, and #516 already shrank the window.

**`celsius_temp_increment` is restored — #523.** Regenerated with `dump_grills.py` rather than hand-edited, and the diff is a pure addition: same 147 grills, control board definitions byte-identical, `celsius_temp_increment` the only key added, zero value changes anywhere else. `PBV30DS` and `PBV30DX` are back to declaring `40..150` by 5. The comment above `_DROPPED_GRILL_FIELDS` now says why it should not be dropped again, since the original comment asserting the opposite is what made it look safe.

Once that lands, rebase and your `celsius_temp_increments` branch stops being hypothetical — `test_set_grill_temperature_prefers_a_declared_celsius_list` will be exercising a path real data can reach.

Two things to fold into the same rebase:

1. The constructor ordering from my last comment — move `celsius_temp_increments` after `json` so the positional signature is unchanged.
2. Nothing else. The rest of the review stands as approved-in-substance; I will approve properly once those two are in.

For completeness on the label question you raised across these PRs: I have gone through and labelled all of them, and #517 is `bug`. You can drop the apology from your PR descriptions — it was never yours to fix, and the release notes are our problem to get right.

trailro:
Rebased onto `2c9591b` and pushed at `3e4eeb0`. Both points are in.

**Constructor ordering — you were right, and I should have caught it myself.** `celsius_temp_increments` now sits after `json`, with a comment saying why it is there rather than beside `temp_increments` where it reads better. Verified positionally rather than by eye:

```
main   : name, control_board, has_lights, has_mpc, min_temp, max_temp, meat_probes, temp_increments, json
branch : name, control_board, has_lights, has_mpc, min_temp, max_temp, meat_probes, temp_increments, json, celsius_temp_increments
```

Identical prefix, the new field strictly appended. 699 tests, ruff and mypy clean.

The rebase conflicted with #521 exactly where predicted — one hunk, the `time`/`monotonic`/`floor` imports — resolved by keeping both.

**Now the correction, and it is to something I said.**

I wrote that PBV30DS and PBV30DX "are being offered setpoints their boards ignore today, in the released integration", and called it a better find than the bug this PR fixes. That was wrong, and I overstated your finding rather than checking it.

Those two models are the only entries in `UNSUPPORTED_MODELS`:

```python
UNSUPPORTED_MODELS = (
    "PBX - test 1",  # Nonstandard data format
    "PBV30DS",       # Nonstandard data format
    "PBV30DX",       # Nonstandard data format
)
```

`get_grills()` checks that **before** the `control_board` branch, so `get_grills(control_board=...)` excludes them too — and that call is the only place `ha-pitboss` sources model names, in `config_flow.py:89`. Neither grill can be selected in the config flow, so no HA install can be configured for one. The wrong-setpoints path is not reachable there at all.

What is true: `get_grill()` bypasses the filter, and `PitBoss.start()` resolves its spec through `get_grill()`, so a library caller who names the model directly does reach it. So #523 is worth having — it just lands for pytboss consumers rather than for HA users.

Which means my `celsius_temp_increments` branch is now real but narrow: reachable by name, for two models pytboss declines to support. `test_set_grill_temperature_prefers_a_declared_celsius_list` is exercising a path real data can reach, as you said — via `get_grill`, not via `get_grills`.

There is a certain coherence to it: both grills are excluded for "nonstandard data format", and a hand-declared Celsius list that does not match the derivation is precisely what nonstandard looks like.

I have no ha-pitboss follow-up to send after all — I offered one in my last comment and it would have been fixing something unreachable. Withdrawn.

And thank you for the labels, and for saying the release notes are yours to get right. Dropping the apology.

dknowles2 (approved):
All three items are in at `3e4eeb0`. 699 tests, ruff and mypy clean locally, CI green.

**The constructor is append-only now.** `Grill`'s positional signature is identical to `main` with `celsius_temp_increments` on the end, so nothing an existing caller passes changes position. The docstring explaining *why* it sits after `json` rather than beside `temp_increments` is the right call — without it, the next person to tidy the field order reintroduces the break, and it is invisible in review.

**The celsius branch is no longer hypothetical.** Now that #523 has landed, I checked the real path rather than the fixture:

```
PBV30DS.celsius_temp_increments -> [40, 45, 50, ... 145, 150]
derived from fahrenheit would be   [37, 40, 43, 46, 48, ... ]
PBV4PS2.celsius_temp_increments -> []   (falls through to derivation)
```

And the difference is observable rather than academic: asking for 44°C snaps to **45** with the declared list and **43** without it. Two degrees, on a real grill, silently. That is the bug this closes for those two models.

The unit inference stays as you had it, per David — an argument would have changed what every existing caller means rather than clarified it.

Merging. Thanks for the rebase and for taking the field-ordering point seriously; it was the kind of thing that would have surfaced as a confusing bug report about missing `json` data months from now.

PR #518, opened 2026-08-02
Give send_command a default timeout and stop it leaking futures

### LINKED ISSUE #510: send_command waits forever by default and leaks its future
```python
async with asyncio.timeout(timeout):   # None unless the caller passes one
    await self._send_prepared_command(cmd)
    return await future
```

No caller inside the library passes a timeout, so if the transport drops after the request goes out the future is never resolved and the caller blocks indefinitely. The entry in `_rpc_futures` is never removed either, so abandoned commands accumulate for the life of the session.

It cost me a frozen coordinator until I made my own transport fail its in-flight futures on disconnect — reaching into `_rpc_futures` and `_lock` to do it, which a third-party transport should not have to.

Branch with a test: [`trailro/pytboss:fix/send-command-timeout-and-future-leak`](https://github.com/trailro/pytboss/tree/fix/send-command-timeout-and-future-leak) — a 30s default plus popping the future in a `finally`, which also covers cancellation. Split out of #506.

trailro:
Fixes #510.

`send_command` defaulted to `timeout=None`, so every caller that didn't pass one waited forever. If the grill goes away mid-command — the WiFi module wedging is the common case — the awaiting task never returns, and its future stays in `_rpc_futures` for the life of the process. Nothing removes it: entries are only popped when a reply with a matching id arrives.

Two changes: a 30s default (`None` still means wait forever, so the escape hatch is intact), and popping the future in a `finally` so timeouts and cancellations clean up after themselves.

The second half has a knock-on benefit for third-party transports. My local RPC transport currently reaches into `self._rpc_futures` and `self._lock` to fail in-flight commands when the socket drops, because nothing in the base class does it. With this, that goes away and a transport only needs `connect` / `disconnect` / `is_connected` / `_send_prepared_command`.

30s is a guess informed by nothing but my own grill; if you'd rather it were 10 or 60, it's a one-line change.

`pytest` (687), `ruff` and `mypy` clean. Label is on you again, sorry — `bug` fits.

### DIFF
--- a/pytboss/transport.py
+++ b/pytboss/transport.py
@@ -16,6 +16,9 @@ from mypy_extensions import DefaultNamedArg
 
 from .exceptions import RPCError
 
+DEFAULT_TIMEOUT = 30.0
+"""Seconds to wait for a reply before giving up."""
+
 
 class RawStateCallback(Protocol):
     """Callback invoked by a `Transport` with raw, unparsed state payloads."""
@@ -81,13 +84,13 @@ class Transport(ABC):
     async def _send_prepared_command(self, cmd: dict) -> None: ...
 
     async def send_command(
-        self, method: str, params: dict, *, timeout: float | None = None
+        self, method: str, params: dict, *, timeout: float | None = DEFAULT_TIMEOUT
     ) -> dict:
         """Sends a comand to the device.
 
         :param method: The method to call.
         :param params: Parameters to send with the command.
-        :param timeout: Timeout for the call.
+        :param timeout: Timeout for the call. Pass `None` to wait forever.
         :raise pytboss.exceptions.RPCError: If the device returns an error response.
         :raise TimeoutError: If `timeout` elapses before a response is received.
         """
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
The future leak is real and the fix is the right one — `_on_command_response` only ever pops on a matching reply, so before this every timed-out or cancelled command left an entry in `_rpc_futures` for the life of the process. Popping in `finally` covers both, and the test asserts the dict is empty afterwards rather than just that the call raised, which is the half that would actually catch a regression.

On the 30s default, I went looking for a caller it could break. Every RPC in `fs.py` and `config.py` is a single round trip — `FS.Get`/`FS.Put` chunk at the call site, so the timeout applies per chunk rather than per transfer — and nothing in the current tree waits on the grill doing slow work inside one command. 30s is comfortable. Worth revisiting if the OTA work in #458 lands, since that is the first plausible caller that could exceed it; it passes its own timeout or gets a bug either way.

Two notes, neither blocking:

- `send_command_without_answer` still defaults to `timeout=None`. The consequence is much smaller — nothing waits on a future — but a reader comparing the two signatures will assume the difference is meaningful. Either give it `DEFAULT_TIMEOUT` too or say why not.
- The `finally` acquires `self._lock`. On the uncontended path `asyncio.Lock.acquire` returns without yielding, so cancellation cannot slip between the `try` and the pop; if it is contended during cancellation the pop can be skipped and the entry survives after all. Narrow enough that I would leave it, but it is the one way the leak can still happen.

The knock-on for third-party transports is a good outcome — a transport needing to reach into `_rpc_futures` and `_lock` to fail in-flight commands was a sign the base class was missing something.

Approving; David merges.

PR #519, opened 2026-08-02
Expose whether a control board converts to celsius itself

### LINKED ISSUE #513: No public way to tell whether a control board converts to celsius itself
`JSFunc.converts_to_celsius()` from #495 lives in `tests/test_grills.py`, so consumers cannot reach it.

It matters: on the boards whose routine contains `ftoc`, the frame is Fahrenheit and the JS converts, while on the others the frame is already in the panel's unit. A consumer deciding what unit a value is in — or what unit to send — has no supported way to distinguish them, short of sniffing `_temperatures_js_func` for the string.

Sniffing is also easy to get wrong, which I demonstrated on myself: a naive `'ftoc' in js` counts **12** boards, but the answer is **11**, because PBL2 ships PBL3's conversion block entirely commented out. Your test helper strips comments first for exactly that reason.

Branch with tests: [`trailro/pytboss:feat/expose-converts-to-celsius`](https://github.com/trailro/pytboss/tree/feat/expose-converts-to-celsius) — exposed per routine on `ControlBoard`, since PBL3 converts in its temperatures reply but not its status one. Split out of #506.

trailro:
Fixes #513.

Most boards convert internally and report whichever unit the grill is set to; some report Fahrenheit always and convert with an `ftoc()` helper in their JS. A consumer deciding what unit a reading is in has no way to ask, so it ends up reimplementing the check against `_temperatures_js_func` — which is private, and easy to get wrong.

This adds `converts_temperatures_to_celsius` and `converts_status_to_celsius` on `ControlBoard`.

Two reasons they're separate properties rather than one:

**The two routines for a board don't always agree.** PBL3 converts in its temperatures reply but not in its status reply. One flag would have to pick a side and be wrong half the time.

**Only live code counts.** PBL2 ships PBL3's conversion block entirely commented out — that is the whole difference between those two boards. A plain `"ftoc" in js` counts PBL2 and shouldn't:

```
naive  'ftoc' in js : 12  [LFS PBA PBE PBL2 PBL3 PBM PBM2 PBT PBV PBV2 PBVA PBX1]
live code only      : 11  [LFS PBA PBE      PBL3 PBM PBM2 PBT PBV PBV2 PBVA PBX1]
```

I made that exact mistake in the original issue and had to correct myself, which is the best argument I have for the property existing at all.

It's read off the routine rather than a list of board names deliberately: `grills.json` is refreshed from the upstream API, so a definitions update can introduce a converting board without any code change here.

`pytest` (688), `ruff` and `mypy` clean. Label: `enhancement`, if you would.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -258,6 +258,13 @@ class Command:
         return evaljs(_COMMAND_JS_TMPL % self._js_func, args=args)
 
 
+def _live_code(js: str | None) -> str:
+    """The routine with commented-out code removed."""
+    if not js:
+        return ""
+    return re.sub(r"//.*", "", re.sub(r"/\*.*?\*/", "", js, flags=re.DOTALL))
+
+
 @dataclass(frozen=True)
 class ControlBoard:
     """Specifications for a control board connected via UART."""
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Verified the counts independently rather than taking them from the description, and they reproduce exactly: 12 boards match a naive `"ftoc" in js` on `temperature_function`, 11 once comments are stripped, and the single difference is `PBL2`. That is the claim the property exists for, and it holds.

Two things I checked that the description does not mention:

**The status side is a much bigger difference than the temperatures side.** Naive matching finds `ftoc` in the `status_function` of 10 boards; live code finds it in 2 (`LFS`, `PBX1`). The other 8 — `PBA`, `PBE`, `PBM`, `PBM2`, `PBT`, `PBV`, `PBV2`, `PBVA` — carry the whole conversion block inside a `/* ... */`. I went looking at `PBA` expecting to catch `_live_code` eating real code, since its `ftoc` calls sit on ordinary-looking lines; they are inside a block comment that opens 700 characters earlier. So `converts_status_to_celsius` is doing real work on 8 boards, not just documenting the `PBL3` asymmetry.

**The regex is safe on the current catalogue.** The obvious failure mode for stripping `//` with a regex is a `//` inside a string literal — a URL is the usual culprit. After block comments are removed there is no live string literal containing `//` in any routine of any of the 20 boards, so nothing is being truncated. Worth knowing it is a property of today's data rather than of the approach: if a definitions refresh ever introduces one, this fails silently and in the safe-looking direction. Not worth pre-solving.

Reading it off the routine rather than a board-name list is right for the reason you give — `grills.json` is regenerated from upstream, so a name list would go stale without anything failing.

The tests cover the three cases that matter, including both comment syntaxes and the per-routine disagreement. Approving; David merges.

PR #520, opened 2026-08-02
Add set_temperature_unit()

### LINKED ISSUE #514: No public wrapper for set-celsius / set-fahrenheit

trailro:
Fixes #514.

Every board declares `set-celsius` and `set-fahrenheit`, but neither is reachable without `_send_command`. `isFahrenheit` can be observed and not changed, which makes the unit the one grill setting a consumer can see but not touch.

`set_temperature_unit(fahrenheit: bool)`, in the same shape as the other command wrappers. Smallest of the set.

I went with one method taking a bool rather than two methods mirroring the two MCU commands, since callers generally have a unit in hand rather than a branch. Trivial to flip if you prefer the pair.

`pytest` (688), `ruff` and `mypy` clean. Label: `enhancement`.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -219,6 +219,18 @@ class PitBoss:
             raise UnsupportedOperation
         return await self._send_command(cmd, temp)
 
+    async def set_temperature_unit(self, fahrenheit: bool) -> dict:
+        """Switches the unit the grill itself works in.
+
+        This is the grill's own setting -- the one shown on its panel and
+        used by the values it reports -- not a display preference.
+
+        :param fahrenheit: Whether the grill should work in Fahrenheit.
+        """
+        return await self._send_command(
+            "set-fahrenheit" if fahrenheit else "set-celsius"
+        )
+
     async def turn_light_on(self) -> dict:
         """Turns the light on if the grill has a light."""
         if not self.spec.has_lights:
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Checked the premise: all 20 control boards in `grills.json` declare both `set-celsius` and `set-fahrenheit`, with no exceptions. So skipping the `UnsupportedOperation` guard that `set_probe_2_temperature` and `turn_light_on` carry is correct rather than an oversight — worth one line in the docstring saying so, since the asymmetry with its neighbours is the first thing a reader will wonder about.

One method taking a bool is the right shape. The two MCU commands are an implementation detail, and callers do have a unit in hand.

Tiny thing: only one blank line before `test_set_temperature_unit`, where the file uses two everywhere else. Ruff will not catch it — E3 is not in the default rule set — so it needs a human to notice.

Approving; David merges.

PR #521, opened 2026-08-02
Extrapolate the cached uptime instead of re-reading it

### LINKED ISSUE #512: get_uptime re-queries more often than the codec key changes
It caches for 5 seconds, but `timed_key` buckets uptime into 10-second windows and the firmware accepts both `x` and `x+1`. So with a password set, every authenticated command pays for an extra round trip that changes nothing.

On a polling transport that is measurable: a local RPC round trip is ~370 ms on my grill, and each poll cycle issues several authenticated commands.

Uptime advances with the wall clock, so one reading stays usable — extrapolating from it with a monotonic clock makes `PB.GetTime` a once-per-session call.

Branch with tests: [`trailro/pytboss:perf/reuse-the-cached-uptime`](https://github.com/trailro/pytboss/tree/perf/reuse-the-cached-uptime), re-reading after ten minutes to bound drift. Split out of #506.

trailro:
Fixes #512.

`get_uptime` re-reads `PB.GetTime` every 5s. It's called on every command, to build the codec key — so a 10s poll costs an extra round trip on each pass, for a value that is knowable without asking: uptime advances with the clock.

This reads it once and extrapolates from a monotonic clock, re-reading every 10 minutes to absorb drift and to notice a reboot. Also switches `time()` to `monotonic()`, which is what the arithmetic wanted all along — the old version would jump if the system clock stepped.

Two things I convinced myself of before proposing it, since this feeds the auth key:

- `timed_key` buckets the value, so small drift is absorbed rather than producing a wrong key.
- A grill reboot resets uptime, and extrapolation would then be wrong until the next re-read. That's the case the 10-minute ceiling is for. A reboot also drops the connection, so in practice the window is smaller than that.

If you'd rather it never extrapolated across a reconnect, clearing `_last_uptime` in `disconnect` would close that gap entirely — happy to add it, I left it out as it seemed like belt and braces.

**Conflict note:** this touches the same region of `api.py` as #517. They're a textual overlap rather than a semantic one; whichever lands second I'll rebase.

`pytest` (688), `ruff` and `mypy` clean. Label: `enhancement`.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -8,14 +8,14 @@ from collections.abc import Awaitable, Callable
 from time import monotonic
 
 from .codec import encode, timed_key
-
-_UPTIME_TTL = 600.0
-"""Seconds before the cached uptime is re-read rather than extrapolated."""
 from .config import Config
 from .exceptions import UnsupportedOperation
 from .fs import FileSystem
 from .grills import Grill, StateDict, get_grill
 from .transport import Transport
+
+_UPTIME_TTL = 600.0
+"""Seconds before the cached uptime is re-read rather than extrapolated."""
 
 _LOGGER = logging.getLogger("pytboss")
 
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -5,9 +5,12 @@ import inspect
 import json
 import logging
 from collections.abc import Awaitable, Callable
-from time import time
+from time import monotonic
 
 from .codec import encode, timed_key
+
+_UPTIME_TTL = 600.0
+"""Seconds before the cached uptime is re-read rather than extrapolated."""
 from .config import Config
 from .exceptions import UnsupportedOperation
 from .fs import FileSystem
@@ -65,7 +68,7 @@ class PitBoss:
         self._vdata_callbacks: list[VDataCallback] = []
         self._state = StateDict()
         self._last_uptime: float | None = None
-        self._last_uptime_check: int | None = None
+        self._last_uptime_check: float = 0.0
 
     def is_connected(self) -> bool:
         """Returns whether we are actively connected to the grill."""
@@ -305,17 +308,18 @@ class PitBoss:
     async def get_uptime(self) -> float:
         """Returns the device's uptime, in seconds.
 
-        Cached for up to 5 seconds between RPCs.
+        Read once and then extrapolated, since uptime advances with the
+        wall clock.
 
         :meta private:
         """
-        now = int(time())
-        if not self._last_uptime_check or now - self._last_uptime_check > 5:
+        now = monotonic()
+        if self._last_uptime is None or now - self._last_uptime_check > _UPTIME_TTL:
             result = await self._conn.send_command("PB.GetTime", {})
             self._last_uptime = result.get("time", 0.0)
             self._last_uptime_check = now
-        assert self._last_uptime is not None
-        return self._last_uptime
+            return self._last_uptime
+        return self._last_uptime + (now - self._last_uptime_check)
 
     async def ping(self, timeout: float | None = None) -> dict:
         """Pings the device.
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -14,8 +14,15 @@ from .fs import FileSystem
 from .grills import Grill, StateDict, get_grill
 from .transport import Transport
 
-_UPTIME_TTL = 600.0
-"""Seconds before the cached uptime is re-read rather than extrapolated."""
+_UPTIME_TTL = 60.0
+"""Seconds before the cached uptime is re-read rather than extrapolated.
+
+Kept short because a grill that restarts resets its uptime, and no transport
+routes its reconnect through `PitBoss` -- `ble` and `wss` both reconnect
+internally -- so nothing invalidates the cache when that happens. Until the
+re-read, `timed_key` would be built from an uptime minutes too high, and every
+authenticated command on a password-protected grill would be rejected. This
+bounds that window to a minute while still saving most of the round trips."""
 
 _LOGGER = logging.getLogger("pytboss")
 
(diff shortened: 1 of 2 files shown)


dknowles2 (changes requested):
The idea is right and the arithmetic is right — uptime does advance with the clock, `monotonic()` is what the old `time()` should always have been, and killing a round trip per command on a 10s poll is worth having. One thing needs closing first, and it is the one you already spotted; I think you underrated it.

**The reboot window is a real outage, and clearing `_last_uptime` in `disconnect` would not close it.**

You offered that as belt and braces. The problem is that neither transport routes a reconnect through `PitBoss` at all. `ble.py` reconnects internally behind `_reconnecting`, and `wss.py` reconnects with exponential backoff in a background task; in both cases `PitBoss.stop()` is never called, so `_last_uptime` and `_last_uptime_check` survive untouched across the reconnect. A grill that is switched off and on — the single most common thing an owner does to a grill — comes back with uptime near zero while we keep extrapolating from the old value.

That matters because the value feeds `timed_key`, which buckets at 10 seconds. An extrapolation that is minutes off does not land in a neighbouring bucket, it lands somewhere unrelated, so `_authenticate` produces a wrong `psw` and **every** authenticated command is rejected for up to the full `_UPTIME_TTL`. Before this PR the 5s cache meant the same window existed but self-corrected almost immediately; this widens it to 10 minutes.

Blast radius is limited to grills with a password set — `_authenticate` only attaches `psw` when `self._password` is truthy — so an unprotected grill is unaffected. That bounds it, but a password is exactly the configuration where a 10-minute dead integration is hardest to diagnose.

The cheapest fix that actually covers the auto-reconnect path is to notice the reset rather than to hook a connect event that does not exist: on each read, if the value comes back lower than what we last extrapolated, the grill restarted. That is self-correcting but only at TTL granularity, which is the same 10 minutes.

So I would rather have an explicit signal. Either give `Transport` a connect callback alongside `set_state_callback` and clear the cache from it, or — much simpler, if you would rather not touch the transport interface — drop `_UPTIME_TTL` to something like 60s. Still 10 commands' worth of round trips saved on a 10s poll, and it bounds the failure to a minute. I would take the second happily and leave the first for whenever a transport-level connect hook is wanted for other reasons.

The rest checked out: `timed_key` buckets at 10s and monotonic drift over even 10 minutes is milliseconds, so extrapolation accuracy itself is not a concern. The first-call path is right — `_last_uptime is None` rather than the old falsy check on the timestamp, which means an uptime of 0.0 no longer forces a re-read every time.

On the conflict with #517: agreed, textual not semantic. Either order is fine by me.

trailro:
You were right that I underrated it, and the part I got wrong is the part I offered as the fix.

I checked your reconnect claim rather than take it, since the whole severity argument rests on it:

- `ble.py` sets `_reconnecting = True` around its internal reconnect and only fires `_disconnect_callback` when that flag is false, so a reconnect is invisible from above.
- `wss.py` reconnects inside the `_subscribe` background task with exponential backoff; `connect()` is not re-entered.
- `_last_uptime` and `_last_uptime_check` are assigned in exactly two places: `__init__` and `get_uptime`. Nothing else touches them, and `PitBoss.stop()` is not on either reconnect path.

So clearing the cache in `disconnect` would have done nothing at all for the case it was meant to cover. It was belt and braces that fastened to neither.

I also confirmed the consequence rather than assume it. `timed_key` does `floor(max(uptime - 5, 0) / 10)` — a 10-second bucket, and that `n` is then folded through the key derivation, so being minutes out is not a near miss, it produces an unrelated key. And `_authenticate` only attaches `psw` when `self._password` is truthy, so your blast-radius bound is right: unprotected grills never notice.

**Taken your second option: `_UPTIME_TTL` is now 60s** (`f9dbeab`). I did not take the transport connect callback, for the reason you gave — it is a public interface change that should be motivated by wanting a connect hook, not by patching around one missing here.

I left out the "value came back lower, so it restarted" check as well. Once the TTL is the only thing that triggers a read, that check can only fire at the moment we would already be correcting ourselves, so it adds a branch and no coverage.

One thing that came out of it beyond the fix: `test_get_uptime_is_extrapolated_between_reads` ticked a hardcoded 331s, which is under a 600s TTL and over a 60s one, so lowering the TTL turned it into a test of the re-read path that still passed its own assertion — it failed only because the fake grill's counter moved. It now ticks fractions of `_UPTIME_TTL`, so it cannot quietly change meaning if the constant moves again.

688 tests, ruff and mypy clean.

Worth adding, since it changes how I read the severity: this lands on me directly. My grill has no password today, which is the configuration you identify as immune — and setting one is on my own list. So this would have been a ten-minute dead integration arriving *because* I improved my security posture, with nothing connecting the two.

dknowles2 (approved):
Good at `f9dbeab`. 688 tests pass locally, CI is green, and the 60s TTL is the option I would have picked.

Agreed on both things you declined. A connect callback on `Transport` should be added because someone wants a connect hook, not as a patch for this; and the value-came-back-lower check really can only fire at the moment the TTL is already correcting us, so it would be a branch with no coverage behind it.

The docstring on `_UPTIME_TTL` is the right place for that reasoning — it names the constraint that sets the number, so the next person to look at 60s and think it is conservative will find out why before changing it.

**The test-fragility catch is the better find here.** A test that ticks a hardcoded 331s silently becomes a test of the other branch the moment the TTL drops below it, and it keeps passing on its own assertion the whole time — that is precisely the failure mode this repo's REVIEW.md warns about, and you found it by lowering a constant rather than by looking for it. Expressing the ticks as fractions of `_UPTIME_TTL` fixes it properly: 7.5 + 15 + 30 = 52.5s, comfortably inside the window at any TTL.

On the last point — yes. A ten-minute dead integration arriving as a consequence of setting a password, with nothing in the logs connecting the two, is the shape of bug that costs someone an evening. Worth stating for the record that this was caught before it shipped and not after.

Approving; David merges.

PR #524, opened 2026-08-02
Carry the error code on RPCError and raise Unauthorized for 401

### LINKED ISSUE #515: RPCError does not carry the error code
`RPCError` keeps only the message, so the 401 the firmware returns is indistinguishable from any other RPC error without matching on the string:

```python
return "unauthorized" in str(ex).lower()
```

which is what I am currently doing downstream, and clearly not something to ship.

It matters because on a local connection `PB.GetState` is itself password-gated, so a stale password takes an integration down entirely rather than just breaking writes — and the right response to that is to prompt for the password, not to retry forever.

Either a dedicated exception type or preserving `error.code` on `RPCError` would do it. No branch: it touches the public exception surface and I would rather not guess which you prefer. Split out of #506.

trailro:
Fixes #515.

You said either a dedicated exception type or preserving `error.code` would do it, and that you would rather not have me guess. It turns out the choice is smaller than it looked, because **`Unauthorized` already exists** — it is just never raised from the RPC path. `auth.py` raises it for rejected account credentials; `_on_command_response` throws a bare `RPCError` for everything, including the firmware's 401.

So this does both, without adding a public class:

- `RPCError` gains an optional `code`, so any error code is available rather than only this one.
- The 401 raises the existing `Unauthorized`, which now subclasses `RPCError`.

```python
except Unauthorized:          # the password was rejected
except RPCError as ex:        # still catches the above; ex.code is there
```

**On backwards compatibility**, since this is the public exception surface:

- `RPCError("boom")` still works; `code` defaults to `None`.
- `Unauthorized` moving from `Error` to `RPCError` only widens what catches it. Anything catching `Error`, or `Unauthorized` itself, is unaffected.
- The one genuine break was mine and mypy caught it: I first made `message` required, which breaks `raise Unauthorized` with no arguments — which your own fake transport does, at `tests/test_api.py:104`. `message` now defaults to `""`, so bare `Unauthorized()` behaves exactly as before.

The widening is the one thing worth your judgement: a caller doing `except RPCError` around `auth.login` would now catch a credentials rejection it previously did not. Different call path, so I think it is theoretical, but it is a real change and I would rather point at it than let you find it.

**Verified the tests fail for the right reason** rather than assuming: with `_rpc_error` reverted to the old single-line `RPCError(...)`, `test_a_rejected_password_raises_unauthorized` and `test_other_rpc_errors_keep_their_code` both fail; with it back, all three pass. I did that on a copy of the file rather than via `git checkout`, having destroyed four changes that way on a downstream PR earlier today.

702 tests, ruff and mypy clean.

**One unrelated thing, which is mine to own.** The second commit restores a blank line in `tests/test_api.py` that `ruff format` wants. It is already on `main`: it came from my conflict resolution on #520, where I merged two test functions and dropped the separator. `build-and-test.yml` runs `ruff check` but not `ruff format --check`, which is why it went in green. Happy to split it out if you would rather this PR were single-purpose — and if you want, adding `ruff format --check` to the workflow would stop the next one, though I did not want to change your CI uninvited.

Label: `enhancement`, or `bug` if you read the 401 being indistinguishable as a defect. Still cannot set one myself.

### DIFF
--- a/pytboss/exceptions.py
+++ b/pytboss/exceptions.py
@@ -6,7 +6,16 @@ class Error(Exception):
 
 
 class RPCError(Error):
-    """Raised when an RPC returns an error."""
+    """Raised when an RPC returns an error.
+
+    :param message: The error message returned by the device.
+    :param code: The error code returned by the device, when it sent one.
+    """
+
+    def __init__(self, message: str = "", code: int | None = None) -> None:
+        super().__init__(message)
+        self.message = message
+        self.code = code
 
 
 class InvalidGrill(Error):
@@ -21,8 +30,14 @@ class NotConnectedError(Error):
     """Raised when there is no active connection to use."""
 
 
-class Unauthorized(Error):
-    """Raised when an RPC is not authorized."""
+class Unauthorized(RPCError):
+    """Raised when a request is not authorized.
+
+    Subclasses `RPCError` so that callers already catching that keep catching
+    this. It is also raised by `pytboss.auth` for rejected account
+    credentials, which is not an RPC -- hence the plain `Error` message
+    signature still working.
+    """
 
 
 class UnsupportedOperation(Error):
--- a/pytboss/transport.py
+++ b/pytboss/transport.py
@@ -14,7 +14,7 @@ from typing import Any, Protocol, Self
 
 from mypy_extensions import DefaultNamedArg
 
-from .exceptions import RPCError
+from .exceptions import RPCError, Unauthorized
 
 DEFAULT_TIMEOUT = 30.0
 """Seconds to wait for a reply before giving up."""
@@ -36,6 +36,19 @@ SendCommandFn = Callable[
     Awaitable[dict[Any, Any] | None],
 ]
 """Signature shared by `Transport.send_command` and `send_command_without_answer`."""
+
+
+UNAUTHORIZED_CODE = 401
+"""The code the firmware answers with when the password is wrong or missing."""
+
+
+def _rpc_error(error: dict) -> RPCError:
+    """Build the exception for an error payload from the device."""
+    message = error.get("message", "Unknown error")
+    code = error.get("code")
+    if code == UNAUTHORIZED_CODE:
+        return Unauthorized(message, code)
+    return RPCError(message, code)
 
 
 class Transport(ABC):
@@ -92,6 +105,8 @@ class Transport(ABC):
         :param params: Parameters to send with the command.
         :param timeout: Timeout for the call. Pass `None` to wait forever.
         :raise pytboss.exceptions.RPCError: If the device returns an error response.
+        :raise pytboss.exceptions.Unauthorized: If the device rejects the
+            password. This is an `RPCError`, so catching that still works.
         :raise TimeoutError: If `timeout` elapses before a response is received.
         """
         cmd = await self._prepare_command(method, params)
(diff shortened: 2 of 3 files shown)


dknowles2 (approved):
Verified rather than taken, including the part you said you checked:

* **The negative test reproduces.** I reverted `_rpc_error` to the old single-line `RPCError(...)` on a copy and ran the suite: `test_a_rejected_password_raises_unauthorized` and `test_other_rpc_errors_keep_their_code` both fail, and pass again with it back. So the tests are pinning the behaviour rather than passing alongside it.
* **The bare `raise Unauthorized` is real** — `tests/test_api.py:104`, in the fake transport. Defaulting `message` to `""` is the right fix; making it required would have been a genuine break in the public exception surface, and mypy catching it is the system working.
* **The widening has no downstream victim.** `ha-pitboss` imports `RPCError` and never `Unauthorized`, and catches it in exactly one place — `coordinator.py:142`, converting to `UpdateFailed`. A password rejection already arrived there as a bare `RPCError`, so it lands identically. Nothing catches `Unauthorized` first, so there is no ordering hazard either.
* 702 tests, `ruff`, `ruff format` and `mypy` all clean.

Finding that `Unauthorized` already existed and was simply never raised from the RPC path is the right result — it makes this smaller than the issue implied, and `except Unauthorized` / `except RPCError as ex: ex.code` covers both shapes without a new public class.

On the theoretical widening around `auth.login`: agreed it is theoretical, and pointing at it beats me finding it.

**The blank line is welcome, not scope creep.** I flagged that exact missing separator reviewing #520 and approved over it as cosmetic, so it is half mine. No need to split it out.

**On adding `ruff format --check` to CI** — worth doing, and I would rather it were a deliberate decision of David's than a rider on a bug fix, so leave it out of this one. I will raise it with him.

Labelled `enhancement`. You genuinely cannot set them; that is ours to get right, not something to apologise for. Merging.

dknowles2:
Correcting something I said in [my review here](https://github.com/dknowles2/pytboss/pull/524#pullrequestreview-4839577224), because it was wrong and it is the kind of wrong that misleads.

I wrote that removing the pragma outright was right because *"BLE001 is not in ruff's default rule set, so there was nothing to suppress in either dialect."* **BLE001 is in ruff 0.16's default rule set.** I found out taking you up on the CI suggestion: writing ruff's *documented* default of `["E4", "E7", "E9", "F"]` into a config produced a lint error `main` does not have, because the real default set is much wider — `BLE`, `B`, `S`, `ASYNC` and more. Measured:

```
def f():
    try: pass
    except Exception as ex: print(ex)
```

with no `[tool.ruff]` at all → `BLE001 Do not catch blind exception: Exception`.

**Your change was still correct, for a different reason.** BLE001 exempts a handler that re-raises, and `services.py` does `raise HomeAssistantError(...) from ex`. So no suppression was needed there — but it was needed because of what the code does, not because the rule was off. If that handler had swallowed instead, removing the pragma would have failed CI.

The `# noqa: BLE001` comments in ha-pitboss are load-bearing, not decorative. I had implied otherwise.

**And the CI change is up: #531 here, [ha-pitboss#343](https://github.com/dknowles2/ha-pitboss/pull/343) downstream.** Both add `ruff format --check` and state the config. Two things came out of it worth having:

- Neither repo pinned `line-length` or `target-version`. In ha-pitboss `target-version` was unset entirely, since with no `pyproject.toml` there is no `requires-python` to infer it from — so version-dependent rules had nothing to work against. Setting it surfaced a real `UP047` in that repo's `conftest.py`.
- Neither sets `select`, deliberately, for the reason above. I nearly shipped that mistake; CI caught it, which is a decent argument for the check existing.

So the blank line you restored was the visible half of something worth fixing properly. Thanks for raising it rather than just patching around it.

PR #525, opened 2026-08-02
Reuse one JS interpreter per thread

### LINKED ISSUE #511: dukpy.evaljs does blocking file I/O inside the event loop
Every parse builds a fresh interpreter, whose init reads its runtime `.js` files from disk. Home Assistant flags it on every state parse:

```
Detected blocking call to open with args ('.../dukpy/jsruntime/process_runtime.js',)
inside the event loop by custom integration 'pitboss'
```

Still present with dukpy 0.6.0. The measured cost is small (~0.5 ms with the files in page cache), so this is about correctness rather than speed.

Caching the interpreter is the obvious fix, but I have not sent one: I cannot tell whether the parse routines ever run off the event loop thread in your intended usage — `get_grill()` already goes through `asyncio.to_thread` in at least one place — and a shared interpreter would not be safe if so. Running the eval in an executor is the other option. This one needs your judgement more than my patch. Split out of #506.

trailro:
Fixes #511.

You held this back because you could not tell whether the parse routines ever run off the event loop thread, and a shared interpreter would not be safe if so. That is the right question, and the answer is that they can: `PitBoss.start` resolves the spec through `asyncio.to_thread(get_grill, ...)`, so parsing is not guaranteed to stay on one thread.

So this keeps the interpreter **thread-local** rather than global. That sidesteps the question entirely instead of answering it: each thread gets its own QuickJS context, so nothing is shared across threads and no assumption about QuickJS thread-safety is needed.

**The cost today, measured rather than estimated.** `dukpy.evaljs()` builds a fresh `JSInterpreter` per call, and `JSInterpreter.__init__` evaluates three runtime shims, each via `open()` — process, console and commonjs. So it is three file reads per eval, not one:

```
20 evals, dukpy.evaljs directly : 60 file reads
20 parses, this branch          :  3 file reads   (once per thread)
```

Parsing one state reply evaluates two routines, so a poll costs six reads on the event loop. Downstream that is about to get worse rather than better: adaptive polling at 10s instead of 30s triples the rate.

**What I checked before proposing reuse**, since "cache it" is only safe if the interpreter is actually reusable:

- **No state leaks between evals.** 200 sequential evals return identical results, and alternating the status and temperatures routines does not perturb either. The routines are self-contained functions evaluated as global scripts.
- **Reuse matches fresh.** Byte-identical results against a per-call interpreter.
- **A failed evaluation does not poison it.** Both a thrown `Error` and a syntax error raise `JSRuntimeError` and leave the interpreter working — I checked this because one malformed frame silently breaking every subsequent parse is the failure mode that would make caching a bad trade.
- **Concurrent threads agree.** Four threads parsing the same message get the same answer.

There are tests for each of those, and I verified the reuse test fails on the old behaviour rather than assuming it would.

The import moves from `dukpy.evaljs` to `dukpy.evaljs.JSInterpreter`; nothing public changes.

**The alternative I did not take** was running the eval in an executor. It leaves the per-call construction cost in place, and it would make `parse_status` async, which is a public API change for something that is otherwise a pure function. Happy to switch if you would rather have that.

One thing this does not do: the interpreter is never released, so a thread that parses once holds a QuickJS context for its lifetime. With `to_thread` that is a pool thread, so the count is bounded by the pool. I judged that cheaper than the reads, but say the word if you would rather it were a `WeakValueDictionary` or explicitly disposed.

703 tests, ruff, ruff format and mypy clean. Label: `enhancement` — still cannot set one myself.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -2,15 +2,40 @@
 
 import json
 import re
+import threading
 from collections.abc import Container, Iterable
 from dataclasses import dataclass, field
 from functools import cache
 from importlib import resources
 from typing import Any, TypedDict, cast
 
-from dukpy import evaljs
+from dukpy.evaljs import JSInterpreter
 
 from .exceptions import InvalidGrill
+
+_INTERPRETERS = threading.local()
+"""One JS interpreter per thread.
+
+`dukpy.evaljs()` builds a fresh interpreter for every call, and each one reads
+three runtime `.js` files from disk. Parsing a single state reply evaluates two
+routines, so a poll costs six file reads -- which an asyncio caller does on its
+event loop.
+
+Reusing one is safe here: the routines are self-contained functions evaluated
+as global scripts, and an interpreter survives a failed evaluation. It is kept
+thread-local rather than global because `get_grill()` is called through
+`asyncio.to_thread`, so parsing is not guaranteed to stay on one thread, and a
+QuickJS context is not documented as thread-safe.
+"""
+
+
+def _run_js(code: str, **kwargs: Any) -> Any:
+    """Evaluate `code` on this thread's interpreter."""
+    interpreter = getattr(_INTERPRETERS, "interpreter", None)
+    if interpreter is None:
+        interpreter = JSInterpreter()
+        _INTERPRETERS.interpreter = interpreter
+    return interpreter.evaljs(code, **kwargs)
 
 
 @cache
@@ -255,7 +280,7 @@ class Command:
         if self._js_func is None:
             raise NotImplementedError
 
-        return evaljs(_COMMAND_JS_TMPL % self._js_func, args=args)
+        return _run_js(_COMMAND_JS_TMPL % self._js_func, args=args)
 
 
 def _live_code(js: str | None) -> str:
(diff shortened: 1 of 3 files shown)


dknowles2 (approved):
Measured it rather than took it, and your numbers reproduce exactly. Counting `.js` opens:

```
10 parses on this branch  :  3 reads   (once, on first use)
10 bare dukpy.evaljs calls: 30 reads   (3 per call)
4 threads x 2 parses      : 12 reads   (3 per new thread)
```

The last line is the one that matters — it confirms the thread-local actually isolates rather than just appearing to.

I also checked the failure mode you singled out, because you were right that it is the one that would make this a bad trade. A thrown `Error` and a syntax error both raise `JSRuntimeError` and leave the interpreter working, with results byte-identical to before, and 100 repeat parses stay identical. One malformed frame silently breaking every later parse would have been much worse than six file reads, and it does not happen.

**Thread-local rather than global is the right call, and for the reason you give.** Sidestepping the question is better than answering it: `get_grill` goes through `asyncio.to_thread`, `parse_status` is a public sync function any caller can invoke from anywhere, and QuickJS context thread-safety is not documented. A global would have needed a guarantee the library cannot make.

Declining the executor alternative is also right — making `parse_status` async to fix a file-read cost would be a public API change out of all proportion to the problem.

On the interpreter never being released: agreed it is the cheaper trade. `to_thread` bounds it to the pool, and a `WeakValueDictionary` would not help since the interpreter is exactly what you want kept alive. Worth revisiting only if someone spawns unbounded threads, which is its own problem.

703 tests, ruff, ruff format and mypy clean. Labelled `enhancement`.

PR #527, opened 2026-08-02
Add probe target routing and stop leaking the password into payloads

trailro:
Raised from review of [ha-pitboss#341](https://github.com/dknowles2/ha-pitboss/pull/341), where the reviewer's position was that this belongs here rather than in the integration. Agreed on all of it — the protocol facts are true with no Home Assistant anywhere near them.

Two things: a probe target API that owns the routing, and a password bug this turned up.

## `set_probe_target` / `get_probe_targets`

`set_probe_temperature` and `set_probe_2_temperature` cover the two boards' commands. Across the catalogue that is 42 of 137 models for probe 1 and 26 for probe 2 — counted through `get_grills()`, not the raw JSON, which misses `_COMMAND_SLUG_OVERRIDES` and counts the shadow-variant keys. **No board declares anything for probes 3 or 4.** So a four-probe grill can be given two targets at most, and on 95 models it can be given none.

The vendor's app does not use a command for the rest. It writes them into the control board's virtual data store under `p1T`..`p4T`, and does the comparing itself; the board never acts on those values. `set_probe_target` routes on `probe_target_command()` and falls back to the store, so a caller does not have to know which route a board supports.

Protocol facts now owned here rather than by each consumer:

- keys are `p1T`..`p4T`
- values are **always Fahrenheit**, whatever unit the grill is working in
- the firmware assigns the payload wholesale, so a write must send back everything already there
- writes are rejected unless `moduleIsOn`, and the store is wiped at power-off — hence `UnsupportedOperation` rather than a silent no-op
- board-reported targets win: `p1Target` is reported by all 137 models, `p2Target` by 26

Verified across all seven mJS firmware versions (0.2.3 to 0.6.0), which implement both RPCs identically including the power-off wipe. The ESP-IDF generation ships both RPCs as well, though there I can confirm presence and not behaviour, since it is a compiled binary.

## The password bug

`_authenticate` did `params["psw"] = ...` on the dict it was handed and returned the same object, and `set_virtual_data` passes the caller's dict straight into it. So calling it left the encoded password in the **caller's** payload. Downstream that ended up stored on a coordinator attribute; the review caught it there.

`_authenticate` now returns a copy, and `get_virtual_data` strips `psw` so a read never hands the credential back as if it were data.

**One thing I could not fix, and want to be exact about.** The password still reaches the grill's store. `PB.SetVirtualData` takes authentication in the same params object the firmware assigns wholesale, so an authenticated write necessarily leaves `psw` in the scratchpad. I found this because my own test asserted the payload was clean and failed on the `with_password` fixture — the fake was right and I was wrong. Sending it unauthenticated is not an option on a password-protected grill.

So this is a protocol constraint, not something the library can avoid, and the test says so rather than asserting something untrue. What the library *can* do is make sure no consumer ever sees it or has to know — which is the argument for the routing living here at all.

I verified the mutation test fails on the old `_authenticate` rather than assuming it would.

## Notes

`_from_fahrenheit` rounds rather than flooring. That is deliberate and different from `set_grill_temperature`, which floors to match what the boards do in their own conversion routine: nothing converts the virtual data store, so the only requirement is that a value round-trips, and rounding is what does that. 74 °C → 165 °F → 74 °C.

`get_probe_targets` omits probes with no target rather than mapping them to `None`, so `targets.get(n)` is the natural read.

`set_virtual_data` and `get_virtual_data` keep their `:meta private:` markers. I had claimed downstream that they were public; they are not, and that claim was wrong.

715 tests, ruff, ruff format and mypy clean. The fake transport now keeps a real scratchpad and assigns wholesale, so these exercise the protocol rather than a mock.

Label: `enhancement`, plus `bug` for the `_authenticate` half if you split them that way. Still cannot set one myself.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -430,6 +430,11 @@ class PitBoss:
         verbatim, so an authenticated write leaves the encoded password in the
         scratchpad and it would otherwise be handed back as if it were data.
 
+        This is the only thing that removes it, rather than a belt-and-braces
+        measure. Firmware 0.6.0 attempts the same strip and misses -- it
+        clears `vData.pws` while `checkPassword` reads `params.psw` -- and the
+        eight earlier mJS versions do not attempt it at all.
+
         :meta private:
         """
         data = await self._conn.send_command(
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -34,6 +34,23 @@ VDataCallback = Callable[[dict], Awaitable[None] | None]
 """A callback function that receives updated VData."""
 
 
+def _to_fahrenheit(temp: int, already_fahrenheit: bool) -> int:
+    """Convert a grill-unit temperature for the virtual data store.
+
+    The store is always Fahrenheit, whatever unit the grill is set to.
+    """
+    if already_fahrenheit:
+        return temp
+    return round(temp * 9 / 5 + 32)
+
+
+def _from_fahrenheit(temp: float, want_fahrenheit: bool) -> int:
+    """Convert a virtual data temperature into the grill's own unit."""
+    if want_fahrenheit:
+        return round(temp)
+    return round((temp - 32) * 5 / 9)
+
+
 class PitBoss:
     """API for interacting with PitBoss grills over Bluetooth LE."""
 
@@ -172,11 +189,16 @@ class PitBoss:
                     callback(vdata)
 
     async def _authenticate(self, params: dict) -> dict:
-        if self._password:
-            params["psw"] = encode(
-                self._password, key=timed_key(await self.get_uptime())
-            ).hex()
-        return params
+        """Return `params` with the encoded password added.
+
+        A copy: the caller's dict is left alone. `PB.SetVirtualData` hands its
+        whole params object to the firmware, so mutating in place put the
+        encoded password into the caller's payload as well as on the wire.
+        """
+        if not self._password:
+            return params
+        psw = encode(self._password, key=timed_key(await self.get_uptime())).hex()
+        return {**params, "psw": psw}
 
     async def _send_hex_command(self, cmd: str) -> dict:
         return await self._conn.send_command(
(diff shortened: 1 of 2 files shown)


trailro:
I re-pulled the firmware to re-verify the claims in this PR rather than trust my notes, and it turned up two corrections and one finding that changes the argument in your favour.

Source is the vendor's CDN, which is public and needs no authentication: `https://firmware.dansonscorp.com/<version>/init.js`.

**Correction 1: there are nine mJS versions, not seven.** I said seven here and on ha-pitboss#341. Sweeping `0-{1..7}-{0..9}` finds:

```
0.2.3  0.3.0  0.4.0  0.5.0  0.5.2  0.5.5  0.5.6  0.5.7  0.6.0
```

I had missed 0.5.2 and 0.5.5. Every claim below is checked against all nine.

**Correction 2: the write is not password-gated on the three oldest.** `checkPassword` appears in `PB.SetVirtualData` only from 0.5.0 onward; 0.2.3, 0.3.0 and 0.4.0 accept an unauthenticated virtual-data write. Does not affect this PR — we always authenticate — but "authenticated write" is not universally true of the firmware and I should not have implied it was.

Everything else holds on all nine, verbatim:

| | |
|---|---|
| both RPCs present | 9 of 9 |
| `vData = params` — wholesale assign | 9 of 9 |
| `if (!moduleIsOn) return {error: -1, "Grill is Off"}` | 9 of 9 |
| `if (moduleIsOn === false) { vData = null; }` on status frames | 9 of 9 |
| `PB.GetVirtualData` returns `vData` unfiltered | 9 of 9 |

**And the finding: 0.6.0 tries to strip the credential and misses by a transposition.**

```js
vData = params;
vData.pws = undefined;   // 0.6.0 only
```

`checkPassword` reads `params.psw`. The strip clears **`pws`**. So the newest firmware attempts exactly the mitigation we were discussing, and the typo means the encoded password stays in the store — which is why my test asserting a clean payload failed against the fake, and why it fails against real hardware too.

That changes how I would describe the `get_virtual_data` strip in this PR. I wrote it as a defensive measure. It is not: it is the only thing that works, because the vendor's own attempt is a no-op on every unit in the field, and on the eight versions before 0.6.0 there is no attempt at all.

It also means a grill that has ever had a virtual-data write from a password-protected client is holding that client's encoded password until the next power cycle, retrievable by anyone who can read the store. `PB.GetVirtualData` is itself password-gated from 0.5.0, so it is not open to an unauthenticated reader — but on 0.2.3 through 0.4.0 neither RPC is gated at all.

I am not proposing anything further here; the PR already does what the library can do. I would rather the reasoning in the docstring matched the actual reason, so I have updated it to say the firmware's own strip is broken rather than absent.

The firmware is archived locally now, so this is re-checkable rather than something I am asking you to take on trust. Happy to attach the relevant `init.js` excerpts to an issue if that is useful to have in the repo.

dknowles2 (approved):
This is the shape I was asking for, and it is better than what I described.

**Verified rather than taken:**

* **The mutation test fails on the old `_authenticate`.** I put the in-place version back on a copy and ran it: `test_the_password_never_reaches_the_scratchpad_as_data[with_password]` fails, `[no_password]` still passes, and both pass with the fix. So the test is pinning the bug rather than passing alongside it.
* **The round-trip holds.** `_from_fahrenheit` rounding where `set_grill_temperature` floors looked like an inconsistency until I checked the reason you gave: every Celsius value from 0 to 120 survives a round trip through the store with zero failures. Flooring would not, and since nothing converts the virtual data store there is no board behaviour to match — round-tripping is the only requirement. Right call, and the comment explaining why the two differ is what stops someone "fixing" it later.
* 718 tests, ruff, ruff format and mypy clean.

**The test that records what cannot be fixed is the best thing in this PR.** Asserting the payload reaching the grill is clean, watching it fail on the password fixture, and then writing down *why* it fails instead of deleting the assertion — that is the opposite of the failure mode I have been complaining about all day. The comment explains the protocol constraint at the point where someone would otherwise try to "fix" it.

And it does sharpen the argument rather than weaken it, exactly as you say: if reading virtual data hands back a credential, that is precisely the knowledge that must not be spread across consumers. Stripping on read is the library doing the only thing it can.

The 0.6.0 detail — the firmware clearing `vData.pws` while `checkPassword` reads `params.psw` — I cannot verify from here, since that lives in the device image rather than anything in this repo. Taking it on your word, and noting it in the docstring is the right place for it: a vendor typo is exactly what a future reader would otherwise assume was our bug.

**On the split:** pytboss owning board-reported-over-scratchpad and Home Assistant keeping the restored-copy rung is a better line than the one I drew. Both of the first two are things the grill told us; the third exists only because HA restarts and the grill's store does not survive a power cycle. Same order, same reasoning, divided by which system owns which kind of fact.

And thank you for correcting the private-API claim on the downstream PR unprompted. It would have been easy to let that one go quiet.

Labelled `enhancement`. It carries a real bug fix too, but the `bug` half is the `_authenticate` leak and I would rather the release notes lead with the capability. Merging.

Issue #505, opened 2026-08-02
Local RPC - potential transport layer


Giordano22:
The ESP32 firmware exposes local RPC methods over HTTP at:

```
http://<grill-ip>/rpc
```

This appears to provide the same interface currently used through the Dansons websocket transport.

Tested locally:

- `PB.GetState` returns the expected `sc_11` / `sc_12` status frames.
- `PB.SendMCUCommand` successfully sends MCU commands.

Example payload to turn off the grill:

```json
{
  "id": 1,
  "method": "PB.SendMCUCommand",
  "params": {
    "command": "FE0102FF"
  }
}
```

The existing MCU command mappings and status decoders should be reusable with a local transport.

## Considerations

It is unclear if this is:

- A firmware-specific behavior.
- An intentionally supported service interface.
- A feature that is unavailable on other models/firmware versions.

The firmware includes password-related RPC functionality (`PB.SetPassword`), but my device has no password configured, so the local RPC interface is currently open on the LAN.

Potential benefit: add a local RPC transport option alongside the existing Dansons websocket transport for users who want a fully local integration.

dknowles2:
@trailro mentioned this in https://github.com/dknowles2/ha-pitboss/issues/321. I don't recall seeing open ports on my smoker, so this may not be universally available. I'll try to test it out tomorrow.

Giordano22:
Like trailro, I have a PBL board so it could very well be specific to us.  I did pull /rpc/config.get and I'll attach at the end here the http section.

Reading init.js it seems that any rpccmd over the websocket just gets passed to RPC.LOCAL. if that's the case generally, a config.set and config.save might enable the http server permanently.

One caveat is that I have not verified whether the Dansons websocket path filters RPC methods before they reach the device. However, Mongoose OS RPC is also enabled over BLE on my board, and that transport should expose the same RPC namespace without a cloud intermediary. Since Bluetooth control is already working in this project, it may be another avenue to test configuration changes.

```
"http": {
    "enable": true,
    "listen_addr": "80",
    "document_root": "/",
    "index_files": "",
    "ssl_cert": "",
    "ssl_key": "",
    "ssl_ca_cert": "",
    "upload_acl": "*",
    "hidden_files": "",
    "auth_domain": "",
    "auth_file": ""
  },
```

dknowles2:
Yeah, my smoker is not exposing the HTTP interface locally:

```
"http": {
  "auth_domain": "",
  "auth_file": "",
[...]

PR #528, opened 2026-08-02
Let a transport fail its in-flight commands when the link drops

### LINKED ISSUE #526: Transport cannot fail in-flight commands when the connection drops
`Transport` has no way to fail the commands it is holding when the connection drops, so every transport that can lose its link has to reach into `self._rpc_futures` and `self._lock` to do it.

#518 improved this a lot — before it, a dropped connection left the caller waiting forever. It now waits `DEFAULT_TIMEOUT`. But the wait is still the full 30s for a command whose reply provably cannot arrive, and nothing in the base class shortens it.

**This is not specific to a local transport.** Neither `ble.py` nor `wss.py` does anything with in-flight futures when their connection goes away — `_on_command_response` is the only place a future is ever completed, so a drop mid-command means the caller sits until the timeout on both of them today:

```
$ grep -n "_rpc_futures\|set_exception" pytboss/ble.py pytboss/wss.py
(no matches)
```

For an integration polling every 10s, that is three polls' worth of stall after the link drops, per command in flight.

**Suggested shape** — a protected method on `Transport` that a subclass calls from wherever it notices the drop:

```python
async def _fail_pending_commands(self, ex: Exception | None = None) -> None:
    """Fail commands still waiting for a reply that can no longer arrive."""
    async with self._lock:
        futures = list(self._rpc_futures.values())
        self._rpc_futures.clear()
    for future in futures:
        if not future.done():
            future.set_exception(ex or NotConnectedError())
```

The alternative is for the base `disconnect()` to do it, but the drop is usually noticed by the subclass's receive loop rather than by an explicit `disconnect()` call, so a hook the subclass can call seems the better fit. `ble.py` and `wss.py` could then call it from their own reconnect paths and get the same benefit.

I raise it because it is the last thing standing between the transport in #505 and a version that touches no private API — that one carries the snippet above verbatim today, which is exactly the reaching-in the base class should make unnecessary. But the wrappers are the ones that would benefit most, since they are the shipped ones.

Happy to send it in whichever shape you prefer, including wiring up `ble` and `wss`. Not building it first, since it is your call where the hook belongs.

trailro:
Fixes #526.

Adds `Transport._fail_pending_commands(ex=None)` and calls it from both shipped transports, so a dropped link fails the commands waiting on it instead of leaving them to time out.

**Wired into the two existing transports, not just made available.** That was the part of the issue I cared about — the gap is theirs, not only a local transport's:

- `ble.py` — from `_on_disconnected`. Bleak calls that synchronously, so the cleanup is scheduled on the loop rather than awaited.
- `wss.py` — from `_subscribe`, right where it clears `self._sock` after the socket closes and before it starts backing off. That is the point where the reply provably cannot arrive: the reconnect gets a new socket, and the device does not resend.

`disconnect()` is deliberately not the hook. A drop is usually noticed by the receive loop, not by an explicit disconnect, so hooking `disconnect()` would miss the case that matters.

**Before this**, on both transports, a command in flight when the link went away waited out `DEFAULT_TIMEOUT` — 30s since #518, and forever before it. For a caller polling every 10s that is three polls of stall for a reply that cannot come.

The `ex` parameter is there because the two callers have different information: `wss` knows the socket closed, `ble` knows the adapter dropped, and a future caller may know the grill went away rather than the link. Defaults to `NotConnectedError`, which is what both use today.

**Verified it is the hook doing the work**, not the timeout: with the body replaced by `return`, a caller with `timeout=None` stays blocked past a two-second bound; with it restored, it wakes with `NotConnectedError`. The tests themselves pass `timeout=None` so they cannot pass by accident on a short timeout.

705 tests, ruff, ruff format and mypy clean.

This is the last thing standing between the transport in #505 and one that touches no private API — it carries this exact snippet today, reaching into `_rpc_futures` and `_lock` from outside. But the two shipped transports are the ones that get the behaviour change.

Label: `bug` seems right, since both transports stall today. Still cannot set one myself.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -88,6 +88,8 @@ class BleConnection(Transport):
         """Called when our Bluetooth client is disconnected."""
         _LOGGER.debug("Bluetooth disconnected.")
         self._is_connected = False
+        # Bleak calls this synchronously, so the cleanup has to be scheduled.
+        self._loop.create_task(self._fail_pending_commands())
         if not self._reconnecting and self._disconnect_callback is not None:
             self._disconnect_callback(client)
 
--- a/pytboss/transport.py
+++ b/pytboss/transport.py
@@ -14,7 +14,7 @@ from typing import Any, Protocol, Self
 
 from mypy_extensions import DefaultNamedArg
 
-from .exceptions import RPCError, Unauthorized
+from .exceptions import NotConnectedError, RPCError, Unauthorized
 
 DEFAULT_TIMEOUT = 30.0
 """Seconds to wait for a reply before giving up."""
@@ -122,6 +122,24 @@ class Transport(ABC):
             # behind forever.
             async with self._lock:
                 self._rpc_futures.pop(cmd["id"], None)
+
+    async def _fail_pending_commands(self, ex: Exception | None = None) -> None:
+        """Fail commands still waiting for a reply that can no longer arrive.
+
+        A subclass calls this wherever it notices the link has gone -- usually
+        its receive loop rather than `disconnect()`, since a drop is not
+        normally an explicit disconnect. Without it, a caller waits out the
+        full `timeout` for a reply the device can no longer send.
+
+        :param ex: The exception to fail them with. Defaults to
+            `NotConnectedError`.
+        """
+        async with self._lock:
+            futures = list(self._rpc_futures.values())
+            self._rpc_futures.clear()
+        for future in futures:
+            if not future.done():
+                future.set_exception(ex or NotConnectedError())
 
     async def send_command_without_answer(
         self, method: str, params: dict, *, timeout: float | None = None
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -118,6 +118,7 @@ class WebSocketConnection(Transport):
                 _LOGGER.debug("WebSocket closed")
 
             self._sock = None
+            await self._fail_pending_commands()
         _LOGGER.debug(
             "Exiting subscribe loop. is_running=%s, keep_running=%s",
             self._loop.is_running(),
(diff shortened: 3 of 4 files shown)


dknowles2 (approved):
Reproduced the negative check, and it is more emphatic than "fails": with `_fail_pending_commands` replaced by `return`, `tests/test_transport.py` does not fail, it **hangs** — I killed it at two minutes. Restored, the same six tests pass in 0.03s. Passing `timeout=None` in the tests is what makes that possible, and it is the right call: on a short timeout they would pass whether or not the hook exists.

725 tests, ruff, ruff format and mypy clean.

**Wiring both shipped transports is what makes this worth having**, and I agree with where you put the calls. `wss.py` right after `self._sock = None` is the point where the reply provably cannot arrive — the reconnect gets a new socket and the device does not resend. And declining `disconnect()` as the hook is correct for the reason you give: a drop is noticed by the receive loop, so hooking the explicit path would miss the case that actually happens.

The `ex` parameter earns its place. `wss` knows the socket closed and `ble` knows the adapter dropped; defaulting to `NotConnectedError` keeps both honest without either having to invent detail.

It also composes correctly with #518: `send_command`'s `finally` pop becomes a no-op once the dict is cleared, and the `if not future.done()` guard means a future whose waiter was already cancelled is left alone rather than raising `InvalidStateError`. Those two are easy to get wrong together and they are not.

One small thing, not worth blocking:

```python
self._loop.create_task(self._fail_pending_commands())
```

Nothing holds a reference to that task. asyncio's docs warn that a task with no strong reference can be garbage collected mid-execution. The window is tiny here and the body is short, but the failure would be silent and would look exactly like the bug this fixes. A `self._cleanup_task = self._loop.create_task(...)` costs a line if you are touching the file again.

Labelled `bug` — both shipped transports stall today, so it is a defect rather than a capability.

PR #530, opened 2026-08-02
Wrap Sys.Reboot and PB.WiFiAwakeWDT

### LINKED ISSUE #529: No public wrapper for Sys.Reboot or PB.WiFiAwakeWDT
Two RPCs the firmware exposes have no public wrapper, so an integration has to reach through `_conn` and `_authenticate` to use them:

```python
await api._conn.send_command_without_answer("Sys.Reboot", {})
await api._conn.send_command("PB.WiFiAwakeWDT", await api._authenticate({}))
```

That is what I run today, and it is two private APIs for two one-line calls.

**Why they matter.** Rebooting the WiFi module is the standard fix when it wedges — over four weeks of Bluetooth use mine dropped repeatedly, once for 32 minutes, caused by the ESP module rebooting itself. And the keep-awake watchdog is what stops the module dropping its connection in the first place. Two Home Assistant buttons are waiting on them; they are the last piece of a set that is otherwise finished.

**What I can show about `PB.WiFiAwakeWDT`**, from the vendor's firmware — the CDN at `https://firmware.dansonscorp.com/<version>/init.js` is public and needs no login:

```js
RPC.addHandler("PB.WiFiAwakeWDT", function (params) {
  if (!checkPassword(params)) {
    return { error: 401, message: "Unauthorized" };
  }
  wsWDT = 5 * 60;   // 5 minutes timeout
  wsTime = 0;
  return null;
});
```

Present and identical in **all nine** mJS versions — 0.2.3, 0.3.0, 0.4.0, 0.5.0, 0.5.2, 0.5.5, 0.5.6, 0.5.7, 0.6.0. It is authenticated, and it returns `null` rather than a result, which is worth knowing for the wrapper's signature. The five-minute figure is the firmware's, not a guess.

**`Sys.Reboot` I can be less certain about, and would rather say so.** It does not appear in `init.js` because it is a Mongoose OS core RPC rather than one the app registers — the firmware's own nine `Sys.reboot` calls are the mJS API function, not the RPC. `Sys.GetInfo` from that same core service is verified working on my grill (it backs two diagnostic sensors), so the service is enabled. But I have not sent `Sys.Reboot` itself, and I am not going to test it casually on a grill in use.

**Relationship to #458.** That PR already implements both, and I said so there rather than duplicating it. I am raising this separately because #458 is much larger — OTA, `PB.LoadMCUFirmware`, `PB.SetWifiCredentials`, `Wifi.Scan` — and has been open since June. Those are exactly the RPCs I decided never to expose from an integration, since a wrong call bricks the board or locks the grill off the network. A narrow PR for the two harmless ones may be easier to land than the whole surface.

I have a branch with just the two wrappers and tests. Entirely happy for you to close this and land #458 instead if you would rather have it in one piece — I have no interest in cutting across your own PR, only in not being blocked behind the OTA review.

trailro:
Fixes #529.

Two wrappers, nothing else:

```python
async def reboot(self) -> None: ...          # Sys.Reboot, no answer expected
async def wifi_awake_wdt(self) -> dict: ...  # PB.WiFiAwakeWDT, authenticated
```

**This duplicates part of your own #458, deliberately and narrowly.** #458 implements both alongside OTA, `PB.LoadMCUFirmware`, `PB.SetWifiCredentials` and `Wifi.Scan` — the RPCs where a wrong call bricks the board or locks the grill off its network, and which I decided never to expose from an integration for exactly that reason. Those deserve the careful review they are getting. These two do not, and two Home Assistant buttons are the only thing waiting.

**Close this without ceremony if you would rather land #458 in one piece.** I would rather lose the work than cut across your PR; the issue explains the reasoning and I am happy for it to end there.

**What is verified, and what is not:**

`PB.WiFiAwakeWDT` — present and identical in all nine mJS firmware versions (0.2.3 through 0.6.0), read from the vendor's public CDN. It is password-checked, so the wrapper authenticates, and the five minutes in the docstring is the firmware's own `wsWDT = 5 * 60` rather than an assumption. It returns `null`, so the wrapper returns whatever the transport gives back rather than promising a shape.

`Sys.Reboot` — **not independently verified, and I would rather flag that than let it read as checked.** It is absent from `init.js` because it is a Mongoose OS core RPC rather than one the app registers; the firmware's own `Sys.reboot` calls are the mJS API function, not the RPC method. `Sys.GetInfo` from that same core service works on my grill, so the service is enabled — but I have not sent `Sys.Reboot`, and I am not going to test it casually on a grill someone is cooking on. #458 makes the same call, so if it is wrong it is wrong in both.

`send_command_without_answer` is used for the reboot because the board goes down before it can reply — same as #458 does, and for the same reason.

709 tests, ruff, ruff format and mypy clean.

Label: `enhancement`. Still cannot set one myself.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -363,7 +363,7 @@ class PitBoss:
         """
         await self._conn.send_command_without_answer("Sys.Reboot", {})
 
-    async def request_fast_updates(self) -> dict:
+    async def request_fast_updates(self) -> None:
         """Asks the grill to push status to the cloud faster, for 5 minutes.
 
         Despite the RPC's name (`PB.WiFiAwakeWDT`) this does not keep the WiFi
@@ -375,14 +375,23 @@ class PitBoss:
         the push timer, so the grill pushes immediately and then every 5
         seconds until the five minutes lapse. Calling again restarts them.
 
-        **It affects the cloud WebSocket only.** The push it accelerates is
-        `WS.send(wsConn, ...)`, so a Bluetooth or local connection sees no
-        difference. Useful for a client reading the grill through the Dansons
-        relay while a user is watching; a no-op for anything else.
+        Two conditions decide whether it does anything at all:
+
+        * **The cloud WebSocket only.** The push it accelerates is
+          `WS.send(wsConn, ...)`, so a Bluetooth or local connection sees no
+          difference.
+        * **Only while the grill is on**, or was on at the previous tick --
+          the firmware guards the push with `lastWasOn || moduleIsOn`. On a
+          cold grill the timer still reschedules to the fast interval and
+          nothing is sent.
+
+        Useful for a client reading the grill through the Dansons relay while
+        a user is watching; a no-op for anything else.
+
+        Returns nothing: the firmware handler ends in `return null`, so there
+        is no result to hand back.
         """
-        return await self._conn.send_command(
-            "PB.WiFiAwakeWDT", await self._authenticate({})
-        )
+        await self._conn.send_command("PB.WiFiAwakeWDT", await self._authenticate({}))
 
     async def set_temperature_unit(self, fahrenheit: bool) -> dict:
         """Switches the unit the grill itself works in.
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -363,12 +363,22 @@ class PitBoss:
         """
         await self._conn.send_command_without_answer("Sys.Reboot", {})
 
-    async def wifi_awake_wdt(self) -> dict:
-        """Holds the WiFi module awake for five minutes.
+    async def request_fast_updates(self) -> dict:
+        """Asks the grill to push status to the cloud faster, for 5 minutes.
 
-        The firmware arms a watchdog (`wsWDT = 5 * 60`) that keeps the module
-        from dropping its connection, and each call restarts it. The timeout
-        is the firmware's; it is not configurable.
+        Despite the RPC's name (`PB.WiFiAwakeWDT`) this does not keep the WiFi
+        module awake. The firmware runs a one-second tick that counts `wsWDT`
+        down and, each time its push timer expires, reschedules it to
+        `wsFastInterval` while `wsWDT > 0` and `wsSlowInterval` otherwise --
+        5 and 60 seconds by default, both settable via
+        `PB.SetWiFiUpdateFrequency`. This call sets `wsWDT` to 300 and zeroes
+        the push timer, so the grill pushes immediately and then every 5
+        seconds until the five minutes lapse. Calling again restarts them.
+
+        **It affects the cloud WebSocket only.** The push it accelerates is
+        `WS.send(wsConn, ...)`, so a Bluetooth or local connection sees no
+        difference. Useful for a client reading the grill through the Dansons
+        relay while a user is watching; a no-op for anything else.
         """
         return await self._conn.send_command(
             "PB.WiFiAwakeWDT", await self._authenticate({})
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -354,6 +354,26 @@ class PitBoss:
         """Whether the grill is currently working in Fahrenheit."""
         return self._state.get("isFahrenheit", True) is not False
 
+    async def reboot(self) -> None:
+        """Reboots the WiFi module.
+
+        The standard remedy when the module wedges: it drops its connections
+        and comes back, leaving the grill itself running. No response is
+        expected, since the board reboots before it can answer.
+        """
+        await self._conn.send_command_without_answer("Sys.Reboot", {})
+
+    async def wifi_awake_wdt(self) -> dict:
+        """Holds the WiFi module awake for five minutes.
+
+        The firmware arms a watchdog (`wsWDT = 5 * 60`) that keeps the module
+        from dropping its connection, and each call restarts it. The timeout
+        is the firmware's; it is not configurable.
+        """
+        return await self._conn.send_command(
+            "PB.WiFiAwakeWDT", await self._authenticate({})
+        )
+
     async def set_temperature_unit(self, fahrenheit: bool) -> dict:
         """Switches the unit the grill itself works in.
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -786,3 +786,26 @@ async def test_probe_target_command_reports_the_route(pitboss: api.PitBoss):
     assert pitboss.probe_target_command(1) == "set-probe-1-temperature"
     assert pitboss.probe_target_command(2) is None
     assert pitboss.probe_target_command(4) is None
+
+async def test_reboot():
+    conn = FakeTransport()
+    pitboss = api.PitBoss(conn, "PBV4PS2")
+    await pitboss.start()
+    with mock.patch.object(
+        conn, "send_command_without_answer", AsyncMock(return_value=None)
+    ) as send:
+        assert await pitboss.reboot() is None
+        # Without an answer: the board reboots before it could send one.
+        send.assert_awaited_once_with("Sys.Reboot", {})
+
+
+async def test_wifi_awake_wdt(pitboss: api.PitBoss, conn: FakeTransport):
+    with mock.patch.object(
+        conn, "send_command", AsyncMock(return_value={})
+    ) as send:
+        assert await pitboss.wifi_awake_wdt() == {}
+        method, params = send.await_args.args
+        assert method == "PB.WiFiAwakeWDT"
+        # Authenticated: the firmware checks the password on this one.
+        assert set(params) <= {"psw"}
+        assert ("psw" in params) is bool(pitboss._password)
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -800,10 +800,9 @@ async def test_reboot():
 
 
 async def test_wifi_awake_wdt(pitboss: api.PitBoss, conn: FakeTransport):
-    with mock.patch.object(
-        conn, "send_command", AsyncMock(return_value={})
-    ) as send:
+    with mock.patch.object(conn, "send_command", AsyncMock(return_value={})) as send:
         assert await pitboss.wifi_awake_wdt() == {}
+        assert send.await_args is not None
         method, params = send.await_args.args
         assert method == "PB.WiFiAwakeWDT"
         # Authenticated: the firmware checks the password on this one.
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -800,9 +800,9 @@ async def test_reboot():
         send.assert_awaited_once_with("Sys.Reboot", {})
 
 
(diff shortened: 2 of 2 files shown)


dknowles2 (commented):
Holding at comment rather than approving, because the only open question here is not about the code: **#458 is David's own PR and adds these two methods with the same names and near-identical bodies.** I confirmed that against its diff rather than taking it from your description. Whether to land this subset now or keep #458 whole is his call, and approving would put it in the merge queue as though it were not.

The code itself is right, and one of your two claims I can corroborate from inside this repo.

**`PB.WiFiAwakeWDT` — confirmed here, not just on the CDN.** `fake_firmware/init.js:580` registers it, checks the password (returning 401 if it fails), sets `wsWDT = 5 * 60` and returns `null`. So authenticating is right, the five minutes really is the firmware's own constant rather than an assumption, and returning whatever the transport gives back rather than promising a shape is right too.

**`Sys.Reboot` — your caveat is exactly correct, and the same file shows why.** `init.js:304` calls `Sys.reboot(0)`, which is the mJS API function, not an RPC handler; there is no `RPC.addHandler("Sys.Reboot", ...)` anywhere. So its availability rests on Mongoose's core RPC service being enabled, which `Sys.GetInfo` working demonstrates but does not prove for `Sys.Reboot` specifically. Flagging that rather than letting it read as checked is the right way to have written it up, and not testing it casually on a working grill is the right way to have left it.

725 tests, ruff and mypy clean. `send_command_without_answer` for the reboot is correct for the reason you and #458 both give.

**On the overlap.** Your framing is fair — these two are wrappers around RPCs that cannot brick anything, while #458 also carries OTA, `PB.LoadMCUFirmware` and `PB.SetWifiCredentials`, which can. Splitting the harmless pair out so two Home Assistant buttons are not waiting on a review of the dangerous ones is a reasonable thing to propose. It is still David's to decide, and if he would rather keep #458 in one piece then closing this costs nothing that is not already written down in #529.

I have labelled it `enhancement` either way, so the decision is not also a labelling chore.

dknowles2:
Do we have a better description of what `PB.WiFiAwakeWDT` does? I'm hesitant to expose that exactly as named, because it's not obvious (to me, at least) what it means at first glance.

Otherwise I'm fine with merging your PR over mine.

trailro:
Good instinct — the name is wrong, and not just unclear. **It does not keep the WiFi module awake.** I traced it through the firmware.

The one-second tick in `init.js`:

```js
if (wsWDT > 0) { wsWDT--; }
wsTime--;
if (wsTime <= 0) {
  wsTime = wsWDT > 0 ? wsFastInterval : wsSlowInterval;
  if (lastWasOn || moduleIsOn) { sendWSStatus(); }
}
```

and the handler:

```js
wsWDT = 5 * 60;
wsTime = 0;
```

So `wsWDT` is a five-minute countdown, and `wsTime` is the countdown to the next status push. Setting `wsTime = 0` makes the grill push immediately, and while `wsWDT > 0` the push timer reschedules to `wsFastInterval` instead of `wsSlowInterval` — **5 seconds instead of 60**, both defaulting in `init.js` and both settable through `PB.SetWiFiUpdateFrequency`. Calling it again restarts the five minutes.

**It is a push-rate boost, and it only touches the cloud.** `sendWSStatus()` ends in `WS.send(wsConn, ...)`, so it accelerates the Dansons relay connection and nothing else. On Bluetooth or a local connection it is a no-op. My guess at intent: the vendor's app calls it while the user has the app in the foreground, and lets it lapse to the 60-second cadence when they put the phone away.

That last part is the bit I would most want a consumer to know, because it is the difference between "useful" and "does nothing" depending on transport.

**Renamed to `request_fast_updates()`** at `1f47c0e`, with the mechanism and the cloud-only caveat in the docstring rather than in a commit message nobody will read. Happy to take a different name — `boost_status_updates`, `request_fast_cloud_updates`, whatever reads best to you — I just wanted it to stop claiming to do something it does not.

This also demotes it for my own purposes: I built a "Keep WiFi awake" button on the strength of the RPC's name, and my grill runs over Bluetooth and local, so that button has never done anything. I will fix that downstream — worth mentioning because it is exactly the misreading the rename prevents.

`Sys.Reboot` is untouched by this. Thanks for corroborating both claims against `fake_firmware/init.js`; that file being in the repo is more useful than the CDN for anything I want reviewed, and I will read it first next time.

739 tests, ruff, ruff format and mypy clean, rebased on current `main`.

dknowles2 (commented):
The rename is right and the trace behind it is correct. I checked every step of it against `fake_firmware/init.js`, which is this repo's model of the firmware rather than the firmware itself — you have the vendor images and I do not — but on this one the model and your CDN reading agree line for line:

* `wsFastInterval = 5` / `wsSlowInterval = 60` (`init.js:38-39`), both overridable from `Cfg` at `:333-337` and both settable via `PB.SetWiFiUpdateFrequency` at `:551-561`.
* The one-second tick at `:466-473` — `wsWDT--`, then `wsTime--`, then on expiry `wsTime = wsWDT > 0 ? wsFastInterval : wsSlowInterval`.
* The handler at `:580-585` — password-checked, `wsWDT = 5 * 60`, `wsTime = 0`, `return null`.
* `sendWSStatus()` at `:176` guards on `wsConn !== null` and ends in a `WS.send`, so the cloud-only caveat holds.

One thing the trace turns up that is worth a line in the docstring: the tick only calls `sendWSStatus()` when `lastWasOn || moduleIsOn` (`:472`). So on a grill that is off, this boosts nothing at all — the timer still reschedules to 5 seconds, but no push happens. "Cloud only" and "only while the grill is on or was just on" are both conditions a caller would want to know before building a button on it.

**`request_fast_updates` is a better name than the RPC's, and I would keep it.** It says what the call does rather than what the vendor named it, and the docstring carries the mechanism. I have no better suggestion.

## One defect, and most of it is ours

`request_fast_updates() -> dict` does not return a dict. The handler returns `null`, and since #534 the transport surfaces that as `None` — `transport.py:178` is `future.set_result(payload.get("result"))`. Confirmed against the branch:

```
request_fast_updates() returned: None | declared -> dict
```

The test does not catch it because `AsyncMock(return_value={})` supplies a dict the real grill never sends.

This is mostly our bug rather than yours: `Transport.send_command` is itself annotated `-> dict` (`transport.py:99-101`), and #534 — merged after you opened this — made that annotation wrong for every result-less handler. You inherited it. Your PR description already says the wrapper "returns whatever the transport gives back rather than promising a shape", which is the right instinct; the signature is the one place that still promises one.

Either `-> dict | None`, or drop the return entirely and make it `-> None`, since there is nothing in the reply worth handing back. I would take the second. The base-class annotation is a separate follow-up and I will open it against ourselves rather than ask you to widen the blast radius here.

`Sys.Reboot` I am taking as unverified exactly as you flagged it, and that flag is the right call — a Mongoose core RPC that is absent from `init.js` because the app never registers it, with `Sys.GetInfo` from the same service working on your grill as the only positive signal. Not something to test casually on a grill mid-cook.

739 tests, ruff, ruff format and mypy all green here too. The `password` fixture being parameterized over both `""` and `"password"` means `assert ("psw" in params) is bool(pitboss._password)` genuinely exercises both branches, which is a nicer test than it looks at first read.

One housekeeping note: the comment above cites `1f47c0e` for the rename, but the branch head is `916908d2` — pre-rebase SHA, I assume. No action needed, just flagging it so nobody goes looking for a commit that is not there.

Holding off merging: David's naming question was the open item, you have answered it with a rename he has not seen yet, and the choice of name is his. Once he confirms it and the return type is fixed, this is ready.

trailro:
Both fixed at `0a7eaad`. And the housekeeping note deserves a worse answer than the one you offered.

**`1f47c0e` was not a pre-rebase SHA. It never existed.** I checked the reflog before replying: the rename was `916908d` from the moment it was committed, and the pre-rebase heads were `38d2f36`, `b322fed` and `5cbdad0`. I wrote a plausible-looking hash into a technical comment instead of reading the one my own push had just printed. Thank you for flagging it rather than assuming I meant something sensible — a fabricated identifier is worse than a stale one, because a stale one at least points somewhere real.

**The return type.** Taken your second option: `-> None`, no return. There is nothing in the reply to hand back, and widening to `dict | None` would only push the question onto callers.

The test was the more interesting half. `AsyncMock(return_value={})` was asserting on a reply no grill sends — and swapping it for `return_value=None` immediately broke authentication, because the same mock intercepts the `PB.GetTime` that `_authenticate` needs. So it is now driven through the fake's own handler:

```python
def _wifi_awake_wdt(self, params: dict) -> None:
    """Password-checked, and ends in `return null` like the firmware."""
    self._check_password(params)
    self.fast_updates_requested = True
    return None
```

Which also makes the `with_password` parameterisation mean something concrete. Checked it: a `PitBoss` holding the wrong password raises `Unauthorized` and the flag stays `False`, so the test would fail if the wrapper stopped authenticating.

**The `lastWasOn || moduleIsOn` guard.** Good catch, and it is in `init.js:472` of the vendor image too — same line, same condition. Added to the docstring as a second bullet, because "cloud only" and "only while the grill is on" are both the difference between a useful button and a decorative one. On a cold grill the timer still reschedules to five seconds and nothing is sent.

On the base-class annotation: agreed that is yours rather than mine, and I would rather it were a separate change than smuggled in here. `Transport.send_command -> dict` has been wrong for every result-less handler since #534, and `send_command_without_answer` is unaffected.

`Sys.Reboot` unchanged, still flagged unverified.

739 tests, ruff, ruff format and mypy clean.

dknowles2 (approved):
David has confirmed `request_fast_updates` as the name, so this is good to go.

Both things I raised are resolved, and I re-ran the branch to check rather than taking the description's word: 739 tests, ruff, ruff format and mypy all clean.

* `-> None` rather than widening to `dict | None` — right call, there is nothing in the reply worth handing to a caller.
* Driving the test through the fake's own password-checked handler instead of a mocked `send_command` is a real improvement over what it replaced. The old assertion was pinning `{}`, a reply no grill sends; the new one exercises the `with_password` parameterisation for something concrete.
* The `lastWasOn || moduleIsOn` condition made it into the docstring as a second bullet. "Cloud only" and "only while the grill is on" together are what stop someone building a button that silently does nothing.

On `1f47c0e` — thank you for checking the reflog and correcting it rather than accepting the charitable reading I offered. I assumed a stale SHA because that is the ordinary explanation; you found it was invented. That distinction matters, and volunteering it when the easier path was to let my guess stand is worth more to me than the commit hash was.

`Sys.Reboot` merges as flagged: verified present as a Mongoose core RPC, not verified by sending it, and #458 makes the same call so it is not a new risk. If it is wrong, it is wrong in both.

The `Transport.send_command -> dict` annotation is ours and stays out of this PR as agreed. Updating the branch past #541 and merging once CI is green.

Thanks for this one — it started as a duplicate of #458 and ends as the better version of that slice, mostly because you kept naming what you had not checked.

PR #534, opened 2026-08-02
Treat an RPC response without a result as a successful empty reply

trailro:
## Problem

`Transport._on_command_response` indexes `payload["result"]` unconditionally. Mongoose OS omits the `result` field when an RPC handler returns nothing (e.g. `Sys.Reboot`, and some `Config.*` handlers), so a perfectly valid success reply raises `KeyError` — and it raises on the **receive path**, not at the caller:

- On BLE, the notification handler errors out, the response is lost, and the caller waits out its full 30s timeout.
- On WSS, the exception propagates into the subscribe loop and kills the connection (no further reconnects).

Verified with a minimal repro: a queued `send_command()` answered with `{"id": N}` (no `result`, no `error`) raises `KeyError: 'result'` from `_on_command_response` while the caller keeps waiting.

## Fix

`payload.get("result")` — the future resolves to `None` for a result-less success, which callers that issue fire-and-forget-style commands simply ignore.

## Testing

- New test: `test_response_without_result_resolves_the_command`.

### DIFF
--- a/pytboss/transport.py
+++ b/pytboss/transport.py
@@ -173,5 +173,7 @@ class Transport(ABC):
             if "error" in payload:
                 future.set_exception(_rpc_error(payload["error"]))
             else:
-                future.set_result(payload["result"])
+                # Mongoose OS omits `result` when a handler returns nothing,
+                # so its absence is a successful empty reply, not an error.
+                future.set_result(payload.get("result"))
         return True
--- a/tests/test_transport.py
+++ b/tests/test_transport.py
@@ -39,6 +39,18 @@ async def test_on_command_response_unknown_id():
     assert handled is False
 
 
+async def test_response_without_result_resolves_the_command():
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
Correct, and it is about to matter more than the description says.

`fake_firmware/init.js` returns `null` from several handlers — `PB.SendMCUCommand` at line 509 and `PB.WiFiAwakeWDT` at line 586 among them. So this is not a hypothetical Mongoose behaviour: the repo's own firmware model exercises exactly the shape that raises today. #530, which wraps `PB.WiFiAwakeWDT`, would land straight on it.

The consequences you describe are the right ones to worry about, and the WSS one is the worse half — an exception on the receive path takes the subscribe loop with it, so the connection stops reconnecting rather than merely losing one reply. #536 hardens that path independently, but a valid reply should not have been raising in the first place.

726 tests, ruff, ruff format and mypy clean. Merging.

PR #535, opened 2026-08-02
Dispatch subscriber callbacks without holding the api lock

trailro:
## Problem

`_on_state_received` / `_on_vdata_received` await subscriber callbacks while holding `PitBoss._lock`. The lock is not reentrant, so any subscriber that calls back into the API deadlocks permanently:

- `get_state()` takes `_lock` to update the cached state → deadlock.
- `subscribe_state()` / `subscribe_vdata()` take `_lock` to append → deadlock.

Verified with a repro: a state subscriber that awaits `get_state()` hangs forever (bounded only by an external `wait_for`).

Separately, the dispatch decision used `inspect.iscoroutinefunction()`, which cannot see through `functools.partial` or an object with an `async __call__` — those were called synchronously and their coroutine silently discarded (plus a `RuntimeWarning`).

## Fix

- State and the subscriber list are updated under `_lock`; callbacks are dispatched **outside** it, under a dedicated `_callback_lock` that preserves the existing guarantee that one update's callbacks finish before the next update's start.
- Whether to await is decided from the returned value (`inspect.isawaitable`) rather than from the callback object.

No public API or behavioral change for existing subscribers: same shared `StateDict` instance, same ordering.

## Testing

- New tests: reentrant `get_state()` from a subscriber, reentrant `subscribe_state()`, and a subscriber object with `async __call__`.
- `pytest` (728 passed), `ruff check`, `ruff format --check`, `mypy .` all green at the CI-pinned versions.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -7,6 +7,7 @@ import logging
 from collections.abc import Awaitable, Callable
 from math import floor
 from time import monotonic
+from typing import Any
 
 from .codec import encode, timed_key
 from .config import Config
@@ -51,6 +52,19 @@ def _from_fahrenheit(temp: float, want_fahrenheit: bool) -> int:
     return round((temp - 32) * 5 / 9)
 
 
+async def _invoke(callback: Callable[[Any], Awaitable[None] | None], arg: Any) -> None:
+    """Call a subscriber, awaiting the result when there is one to await.
+
+    Decided from the returned value rather than the callback:
+    `inspect.iscoroutinefunction` cannot see through `functools.partial` or
+    an object with an `async __call__`, and calling those without awaiting
+    silently discards their coroutine.
+    """
+    result = callback(arg)
+    if inspect.isawaitable(result):
+        await result
+
+
 class PitBoss:
     """API for interacting with PitBoss grills over Bluetooth LE."""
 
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
This is a real deadlock and the fix is the right shape. I flagged the same hazard reviewing #516 — a subscriber calling `get_state()` would take a non-reentrant `_lock` already held by the dispatcher — and left it as a note because nothing in either repo did it. Something eventually would have.

Splitting `_lock` from `_callback_lock` is better than the obvious alternatives: it preserves the ordering guarantee that holding `_lock` across dispatch was providing, without making that guarantee cost reentrancy. Snapshotting the callback list under `_lock` is what makes `subscribe_state()` from inside a callback safe too.

**`inspect.isawaitable(result)` over `iscoroutinefunction(callback)` is the better catch of the two.** A `functools.partial` wrapping a coroutine function, or an object with `async __call__`, was being called and its coroutine dropped — a `RuntimeWarning` and a silently skipped subscriber, which is the kind of thing that gets diagnosed as "the callback didn't fire" for an hour. Deciding from the value rather than the object is correct in general, not just here.

728 tests, ruff and mypy clean.

One consequence worth recording rather than fixing here. Dispatch used to run under the same lock as `_state.update()`, so the shared `StateDict` could not change while a subscriber was reading it. It can now: a concurrent `get_state()` can add keys mid-dispatch, and a subscriber iterating the dict would get `RuntimeError: dictionary changed size during iteration`. Narrow — it needs a polling-and-pushing connection and a subscriber that iterates rather than indexes — but #516 made polling-and-pushing the normal case, so it is not purely theoretical.

The existing `# TODO: Send copies of state so subscribers can't modify it` is exactly the fix, and doing it would close this too. I would not fold it in here: it changes the documented contract that every subscriber gets the same instance, which deserves its own change. Worth noting on that TODO that it now has a second reason to exist.

Merging.

PR #536, opened 2026-08-02
Keep the WSS receive loop alive through bad payloads and callbacks

trailro:
## Problems

Four related weaknesses in `WebSocketConnection`, all verified with repros:

1. **Any exception on the receive path permanently kills the connection.** The subscribe loop only catches `GrillUnavailable` from reconnect attempts. A non-JSON frame (`msg.json()` raising), an exception escaping a subscriber callback, or a `KeyError` from response handling ends the task silently: no further reconnects (despite the class docstring promising them), no `_fail_pending_commands()`, and the task's exception surfaces only at `disconnect()`.

2. **A subscriber that sends a command during dispatch deadlocks the transport.** The loop held `_sock_lock` across `_handle_message()` (and therefore across subscriber callbacks), while `_send_prepared_command()` needs the same lock. Repro: a state callback issuing any send blocks the receive loop forever.

3. **`disconnect()` can block for up to 30 seconds.** If the subscribe loop is in its reconnect backoff sleep, nothing interrupts it; `disconnect()` awaits the task for the remainder of the sleep. Measured: exactly the remaining backoff.

4. **`AttributeError` instead of `NotConnectedError`.** The loop clears `_sock` without the lock, so a sender that passed `is_connected()` and then waited on `_sock_lock` could find `_sock is None` and crash on `None.send_json`.

## Fix

- Per-message try/except in the receive loop: malformed frames and subscriber exceptions are logged and skipped; non-TEXT frames are ignored.
- `_sock_lock` is now held only around the actual send; message dispatch runs without it. (Awaiting a full command round-trip *inside* a callback still cannot complete — the response is read by the loop the callback blocks — but it now fails via its own timeout instead of wedging the stream forever; fire-and-forget sends work.)
- The backoff sleep moved into `_backoff_wait()`, interruptible by `disconnect()` via a `_stopping` event. The existing `test_reconnect_backoff` was updated to patch `_backoff_wait` instead of `asyncio.sleep`; its asserted 1→2→4→…→30 progression is unchanged.
- `_send_prepared_command` re-reads `_sock` under the lock and raises `NotConnectedError` when it is gone.

## Testing

- New tests: malformed payload survival, subscriber-exception survival, send-from-callback (deadlock regression), `NotConnectedError` on a vanished socket, and disconnect-during-backoff.
- `pytest` (730 passed), `ruff check`, `ruff format --check`, `mypy .` all green at the CI-pinned versions.

### DIFF
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -3,12 +3,14 @@
 import asyncio
 import logging
 from asyncio import AbstractEventLoop, Event, Lock, Task
+from contextlib import suppress
 from typing import Any
 from uuid import uuid4
 
 from aiohttp import (
     ClientSession,
     ClientWebSocketResponse,
+    WSMsgType,
     WSServerHandshakeError,
 )
 
@@ -54,12 +56,14 @@ class WebSocketConnection(Transport):
         self._app_id = app_id or str(uuid4()).split("-")[-1]
         self._subscribe_task: Task | None = None
         self._subscribed = Event()
+        self._stopping = Event()
         self._keep_running = False
 
     async def connect(self) -> None:
         """Starts the connection to the device."""
         self._sock = await self._ws_connect()
         self._keep_running = True
+        self._stopping.clear()
         self._subscribe_task = self._loop.create_task(self._subscribe())
         await self._subscribed.wait()
 
@@ -71,6 +75,9 @@ class WebSocketConnection(Transport):
         for the background reconnect/subscribe task to finish.
         """
         self._keep_running = False
+        # Wake the subscribe task if it is waiting out a reconnect backoff;
+        # otherwise this call blocks for the remainder of that sleep.
+        self._stopping.set()
         if self._sock:
             await self._sock.close()
         # Only close the session if we created it (not if it was provided externally)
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Four independent problems, each with a fix that addresses the cause rather than the symptom. The first is the one that matters most: the class docstring promises automatic reconnection with exponential backoff, and any exception on the receive path silently revoked that promise for the life of the process. A per-message boundary is the right place to put the guard.

Checking the pieces:

* **`_sock_lock` scope.** Holding it across `_handle_message` meant it was held across subscriber callbacks, and `_send_prepared_command` wants the same lock — so any subscriber that sent anything wedged the receive loop permanently. Narrowing it to the send is correct, and I appreciate the honesty in the description that a *full round trip* awaited inside a callback still cannot complete: the reply is read by the loop the callback is blocking. That it now fails by timeout instead of wedging the stream is the achievable improvement, and saying so beats implying the deadlock class is gone.
* **`_backoff_wait`.** `asyncio.timeout` around `_stopping.wait()` with `suppress(TimeoutError)` is a clean interruptible sleep, and moving it behind a method is what lets the existing `test_reconnect_backoff` keep asserting the 1→2→4→…→30 progression rather than being rewritten around the change.
* **Re-reading `_sock` under the lock.** Right fix, and dropping the now-redundant `is_connected()` pre-check is right too — it was the stale read that produced `AttributeError` on `None.send_json`.
* **`WSMsgType.TEXT` filter.** Sensible; `msg.json()` on a non-text frame was another way into the same fatal path.

It also composes correctly with #528: `_fail_pending_commands()` is still called where the stream ends, so a drop still fails in-flight commands rather than leaving them to time out.

730 tests, ruff and mypy clean — and the suite is visibly slower here (11s against the usual 1.2s), which is the timing tests actually exercising the backoff rather than mocking it away.

Merging.

PR #537, opened 2026-08-02
Decode file content once instead of per 512-byte chunk

trailro:
## Problem

`FileSystem.get_file_content` decodes each 512-byte `FS.Get` chunk as UTF-8 individually. A multi-byte character that happens to fall across a chunk boundary is not valid UTF-8 on either side of the cut, so the read raises:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 511: unexpected end of data
```

Verified with a repro: a file whose 512-byte boundary splits an `é`.

## Fix

Accumulate raw bytes and decode once at the end.

## Testing

- New test: `test_get_file_content_multibyte_split_across_chunks`.

### DIFF
--- a/pytboss/fs.py
+++ b/pytboss/fs.py
@@ -29,15 +29,17 @@ class FileSystem:
         """
         length = 512
         offset = 0
-        content = ""
+        content = bytearray()
         while True:
             resp = await self._conn.send_command(
                 "FS.Get", {"filename": filename, "offset": offset, "len": length}
             )
-            content += b64decode(resp["data"]).decode("utf-8")
+            content += b64decode(resp["data"])
             offset += length
             if resp["left"] == 0:
-                return content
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Plainly right. `b64decode(...).decode("utf-8")` per 512-byte chunk cannot work for any file whose multi-byte characters do not happen to align to the boundary — and 512 is a fixed byte count, so alignment is luck.

Accumulating into a `bytearray` and decoding once is the correct shape rather than a patch: it also makes the failure mode honest, since a genuinely malformed file now raises on the whole content instead of at whichever chunk happened to split.

The test picks the case that matters — a boundary that splits an `é` — rather than asserting on a file that would have passed anyway.

726 tests, ruff, ruff format and mypy clean. Merging.

PR #538, opened 2026-08-02
Document why the uptime cache is stamped before the request

trailro:
## What this is now

Originally this PR moved the uptime cache timestamp after the `PB.GetTime` round trip. Review showed the premise was inverted — see the discussion below — so it is now the change the review suggested instead: keep the pre-request stamp, and record **why** it is correct in a comment plus a regression test, so the "more accurate" version does not get re-proposed.

## Why the pre-request stamp is correct

`checkPassword` accepts a key built from the firmware's current 10-second bucket **or the next one** — `x` or `x + 1`, never `x - 1`. The tolerance is asymmetric on purpose: a client estimate that runs ahead is forgiven, one that runs behind is rejected. The pre-request stamp biases every extrapolation ahead by about one request latency, which is exactly the forgiven direction; stamping after the reply flips the bias behind, where a slow link draws spurious `Unauthorized` errors near bucket boundaries.

Verified against the vendor firmware images, not just `fake_firmware/init.js`: 0.5.6 (`init.js:496`), 0.5.7 (`init.js:489-497`), and 0.6.0 (`init.js:440-448`, via the `PSW` module) all try `x` then `x + 1` and nothing else. 0.2.3 predates password support.

## Testing

- New regression test `test_get_uptime_extrapolation_runs_ahead_not_behind` pins the bias direction.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -457,12 +457,17 @@ class PitBoss:
         if self._last_uptime is None or now - self._last_uptime_check > _UPTIME_TTL:
             result = await self._conn.send_command("PB.GetTime", {})
             self._last_uptime = result.get("time", 0.0)
-            # Stamped after the reply: the grill computed its uptime then,
-            # not when the request went out. Stamping before the round trip
-            # inflates every later extrapolation by the RTT, and timed_key
-            # rounds to 10-second windows, so a slow link (BLE via a proxy)
-            # can push an authenticated command into the wrong window.
-            self._last_uptime_check = monotonic()
+            # Deliberately the pre-request timestamp, though the grill
+            # computed its uptime later, when the request reached it: every
+            # extrapolation therefore runs ahead of the grill by about one
+            # request latency, and ahead is the only direction the firmware
+            # forgives. checkPassword accepts a key built from its current
+            # 10-second bucket or the next one -- x or x + 1, never x - 1 --
+            # which absorbs a client running ahead but rejects one running
+            # behind. Stamping after the reply reads as more accurate and
+            # is not: it flips the bias behind, where a slow link draws
+            # spurious Unauthorized errors near bucket boundaries.
+            self._last_uptime_check = now
             return self._last_uptime
         return self._last_uptime + (now - self._last_uptime_check)
 
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -457,7 +457,12 @@ class PitBoss:
         if self._last_uptime is None or now - self._last_uptime_check > _UPTIME_TTL:
             result = await self._conn.send_command("PB.GetTime", {})
             self._last_uptime = result.get("time", 0.0)
-            self._last_uptime_check = now
+            # Stamped after the reply: the grill computed its uptime then,
+            # not when the request went out. Stamping before the round trip
+            # inflates every later extrapolation by the RTT, and timed_key
+            # rounds to 10-second windows, so a slow link (BLE via a proxy)
+            # can push an authenticated command into the wrong window.
+            self._last_uptime_check = monotonic()
             return self._last_uptime
         return self._last_uptime + (now - self._last_uptime_check)
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -404,11 +404,14 @@ async def test_get_uptime_is_extrapolated_between_reads(pitboss: api.PitBoss):
         assert await pitboss.get_uptime() == first + sum(steps)
 
 
-async def test_get_uptime_extrapolation_excludes_the_rpc_round_trip():
-    """The grill computes its uptime when it replies, not when we ask.
+async def test_get_uptime_extrapolation_runs_ahead_not_behind():
+    """The RPC round trip biases extrapolations ahead of the grill, on purpose.
 
-    The cache timestamp therefore has to be taken after the round trip;
-    taking it before inflates every extrapolation by the RTT.
+    The firmware's checkPassword accepts a key built from its current
+    10-second bucket or the next one (x or x + 1, never x - 1), so an
+    estimate that runs ahead is forgiven and one that runs behind is
+    rejected. The cache timestamp is therefore taken before the request
+    goes out; stamping it after the reply would flip the bias behind.
     """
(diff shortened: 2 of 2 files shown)


dknowles2 (changes requested):
Please hold this one. I think the premise is inverted, and if it is, this makes the failure it describes **more** likely rather than less — on exactly the password-protected, slow-link grills it names.

## The firmware accepts a client that is ahead, not one that is behind

`fake_firmware/init.js:482-498`:

```js
let x = getCodecTime();                 // floor(max(Sys.uptime() - 5, 0) / 10), at handling time
let key = getCodecKey(KEY, x);
if (codec(data, key, 0) === grillPassword) return true;
key = getCodecKey(KEY, x + 1);          // and also x + 1
return codec(data, key, 0) === grillPassword;
```

Two things follow. The grill computes its bucket from its **own uptime when it handles the command**, not when it answered `PB.GetTime`. And it accepts a key built from `x` **or `x + 1`** — there is no `x - 1`.

So the tolerance is asymmetric by design: a client whose estimate is up to one bucket **high** is accepted; a client whose estimate is **low** is rejected. That asymmetry is not incidental. Our command reaches the grill *after* we compute the key, so the grill's uptime at handling is always a little greater than ours at sending — `x + 1` is what absorbs that.

## Which makes the current code right and this change harmful

* Stamping **before** the round trip biases the estimate **high**, by the request latency.
* Stamping **after** the reply biases it **low**, by the response latency.

I simulated both against the firmware's actual acceptance rule, sweeping uptime across 2000 bucket positions at 1s one-way latency:

```
client biased HIGH (stamp before, today):  accepted 2000/2000  (100.0%)
client biased LOW  (stamp after, this PR): accepted 1620/2000  ( 81.0%)
```

So on the slow link this PR is aimed at, it would introduce roughly a one-in-five chance of `Unauthorized` per authenticated command, where today there is none.

## Where the description goes wrong

> every later extrapolation is inflated by the round-trip time

It is inflated by the **request** latency — the send-to-handling leg — not the full RTT. And inflation is the direction the firmware forgives. Stamping after does not remove the error, it flips its sign into the direction that is not forgiven.

The test passes because the fake computes its `time` value at handling and the simulated 5s all falls before the reply is observed, so "after" looks like it excludes the delay. It is measuring the fake's timing rather than the acceptance rule.

## What I would take instead

If the round trip is worth correcting for at all, the midpoint is the unbiased estimate:

```python
before = monotonic()
result = await self._conn.send_command("PB.GetTime", {})
self._last_uptime = result.get("time", 0.0)
self._last_uptime_check = (before + monotonic()) / 2
```

That halves the error instead of reflecting it. But given the acceptance rule, biased-high is already safe and the current code is not accidentally correct — it is correct for a reason worth writing down. A comment recording *why* the stamp is taken before the send would be worth more than either change, and would stop this being re-proposed.

## One caveat on my own evidence

Everything above rests on `fake_firmware/init.js`, which is this repo's model of the firmware rather than the firmware. It is a good model — `codec.py`'s `timed_key` is documented as a port of the same `getCodecKey` — but you have been reading the real images across nine versions and I have not. **If the real `checkPassword` also tries `x - 1`, my whole argument collapses and this PR is fine.** That is the one thing worth checking before either of us acts on this.

726 tests, ruff, ruff format and mypy are clean, and the change is one line — none of which is in question. It is only the direction.

trailro:
You're right, and the caveat resolves in your favor: I checked the real firmware images, and there is no `x - 1` anywhere.

- **0.5.7** (`init.js:489-497`): `getCodecKey(KEY, x)`, then `getCodecKey(KEY, x + 1)`, nothing else.
- **0.5.6** (`init.js:496`): identical.
- **0.6.0** (`init.js:440-448`): same two attempts via the refactored `PSW` module (`PSW.getTime()` / `PSW.getKey(KEY, x)` / `PSW.getKey(KEY, x + 1)`).
- **0.2.3** predates password support entirely.

So `fake_firmware/init.js` models the real acceptance rule faithfully, the tolerance is asymmetric by design, and the premise of my original change was inverted exactly as you said.

I've reworked the branch into what you suggested taking instead: the stamp stays before the request, with a comment recording why that is load-bearing (asymmetric `x`/`x + 1` window, ahead is forgiven, behind is not) and a regression test that pins the bias direction (`test_get_uptime_extrapolation_runs_ahead_not_behind`), so this doesn't get re-proposed by the next person who reads the pre-request stamp as an oversight. Title and description updated to match. I skipped the midpoint variant — given the acceptance rule, biased-high is already the safe side, and halving the error buys nothing.

dknowles2 (approved):
Thank you for going and checking the real images. That was the one thing that could have settled it and I could not do it myself — I only have `fake_firmware/init.js`, and three vendor versions agreeing that it is `x` then `x + 1` and nothing else is what makes the asymmetry a fact about the firmware rather than a fact about our model of it. I am taking 0.5.6, 0.5.7 and 0.6.0 on your word, and noting that here so the next person knows which part of this is independently checkable in the repo and which part is not.

This is now a better change than either version that came before it. The original moved a line; this leaves the line alone and removes the reason anyone would move it again. Agreed on skipping the midpoint — I offered it as the unbiased estimate, but unbiased is not what you want against a one-sided window, and halving the error toward the rejected side is worse than leaving it entirely on the forgiven side. Dropping it was right.

I checked the regression test does what it says rather than passing incidentally. Reverting the stamp to `monotonic()` after the reply on a scratch copy:

```
E  assert 100.0 >= (100.0 + 5.0)
FAILED tests/test_api.py::test_get_uptime_extrapolation_runs_ahead_not_behind
```

So it fails on exactly the change it exists to prevent, and it is the direction being pinned rather than a specific number. The `_get_time` override lands because `FakeTransport` builds its dispatch per call, and the 5s tick stays well inside `_UPTIME_TTL`, so the second `get_uptime()` is a genuine extrapolation and not a re-read. 726 tests, ruff, ruff format and mypy all green here too.

One bookkeeping note: relabelling this `documentation`. It no longer fixes a bug — it records why there is not one.

The branch is behind `main` by the four PRs that landed after you opened this, so I will update it and merge once CI is green.

PR #541, opened 2026-08-02
Pass the unit flag to temperature commands

trailro:
## Problem

On eleven boards — LFS, PBA, PBE, PBL3, PBM, PBM2, PBT, PBV, PBV2, PBVA, PBX1 — the MCU always speaks Fahrenheit. Two things in the vendor's own routines show it:

1. Their `temperature_function` reports Fahrenheit and converts to Celsius **for display** with `ftoc()` when the frame's unit flag says the panel is in C-mode.
2. Their `set-temperature` (and, on PBA/PBE/PBL3/PBT/PBX1, `set-probe-N-temperature`) routines take a second parameter: `(temp, is_fahrenheit)`, and convert a Celsius argument to Fahrenheit when it is `false`. All 20 such routines in `grills.json` share the identical branch; PBVA even ships a calibrated C→F lookup table (`cval`/`fval`).

pytboss calls these routines with the temperature alone, so `arguments[1]` is `undefined`, the `=== false` branch never fires, and on a grill set to Celsius the setpoint goes through unconverted — **read as Fahrenheit by the MCU**:

| call | wire | grill actually set to |
|---|---|---|
| `set-temperature(100)` — pytboss today, user wants 100 °C | `FE0501 01 00 00 FF` | 100 °F ≈ 38 °C |
| `set-temperature(100, false)` — vendor app, C-mode | `FE0501 02 01 00 FF` | 210 °F ≈ 99 °C |

Probe targets are off the same way: PBL3 `set-probe-1-temperature(70)` sends 70 (°F, ~21 °C) where the app sends 160 °F for 70 °C.

Fahrenheit mode is unaffected (passthrough is the intended path), which is presumably why this has gone unnoticed — it only bites grills whose panel is set to Celsius.

## Fix

`set_grill_temperature`, `set_probe_temperature`, `set_probe_2_temperature` and the board-command path of `set_probe_target` now pass the grill's current unit as the routine's second argument. Safe uniformly: routines without the branch ignore extra arguments, fixed-hex commands ignore arguments entirely, and when no state has arrived yet the flag defaults to `True` (passthrough — exactly today's behavior).

## Testing

- New tests in `test_grills.py` execute the real vendor JS for PBV and PBL3 in both modes and pin the conversion (plus one proving the flag is harmless on PBL, which takes one parameter).
- `test_api.py` assertions updated to expect the flag; new behavior covered by the existing Celsius-mode tests.
- `pytest` (739 passed), `ruff check`, `ruff format --check`, `mypy .` green at the CI-pinned versions.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -262,29 +262,37 @@ class PitBoss:
 
         :param temp: Target grill temperature, in the grill's own unit.
         """
-        fahrenheit = self._state.get("isFahrenheit", True)
+        fahrenheit = self._is_fahrenheit()
         if accepted := self.accepted_setpoints(fahrenheit):
             temp = min(accepted, key=lambda value: abs(value - temp))
-        return await self._send_command("set-temperature", temp)
+        # The unit flag travels with every temperature command: on eleven
+        # boards the MCU always speaks Fahrenheit, and the vendor's command
+        # routine converts a Celsius argument only when its second parameter
+        # is `false`. Sent alone, a Celsius setpoint is read as Fahrenheit
+        # (100C becomes 100F). Boards whose routine takes one parameter
+        # ignore the extra argument, as do fixed-hex commands.
+        return await self._send_command("set-temperature", temp, fahrenheit)
 
     async def set_probe_temperature(self, temp: int) -> dict:
         """Sets the target temperature for probe 1.
 
-        :param temp: Target probe temperature.
+        :param temp: Target probe temperature, in the grill's own unit.
         """
-        return await self._send_command("set-probe-1-temperature", temp)
+        return await self._send_command(
+            "set-probe-1-temperature", temp, self._is_fahrenheit()
+        )
 
     async def set_probe_2_temperature(self, temp: int) -> dict:
         """Sets the target temperature for probe 2.
 
-        :param temp: Target probe temperature.
+        :param temp: Target probe temperature, in the grill's own unit.
         :raise pytboss.exceptions.UnsupportedOperation: When probe 2's
             target temperature cannot be set.
         """
         cmd = "set-probe-2-temperature"
         if cmd not in self.spec.control_board.commands:
             raise UnsupportedOperation
-        return await self._send_command(cmd, temp)
+        return await self._send_command(cmd, temp, self._is_fahrenheit())
 
     def probe_target_command(self, probe_number: int) -> str | None:
         """The board command that sets this probe's target, if it has one.
@@ -314,7 +322,9 @@ class PitBoss:
             clears the store when the grill is switched off.
         """
         if self.probe_target_command(probe_number) is not None:
-            await self._send_command(f"set-probe-{probe_number}-temperature", temp)
+            await self._send_command(
+                f"set-probe-{probe_number}-temperature", temp, self._is_fahrenheit()
+            )
             return
         if not self._state.get("moduleIsOn"):
             raise UnsupportedOperation(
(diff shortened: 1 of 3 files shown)


dknowles2 (approved):
This is a real bug and the fix is the right shape. I reproduced the whole thing rather than reading it, by executing the actual vendor routines out of `grills.json` for every board:

```
BOARDS WHERE THE FLAG CHANGES THE WIRE BYTES:
  LFS   set-temperature          (100)=FE0501010000FF  (100,False)=FE0501020100FF
  PBA   set-temperature/probe1/2 ...same shape, FE0501/0502/0503
  PBE   set-temperature/probe1/2
  PBL3  set-temperature/probe1
  PBM   set-temperature
  PBM2  set-temperature
  PBT   set-temperature/probe1/2
  PBV   set-temperature
  PBV2  set-temperature
```

and confirmed the other half of the safety argument the same way — the eight boards whose routines take one parameter (`LBL`, `PBB`, `PBC`, `PBC2`, `PBD`, `PBG`, `PBL`, `PBP`) return byte-identical output with and without the flag, and a fixed-hex command ignores arguments entirely (`LBL/set-celsius` → `FE0902FF` either way). So the "safe uniformly" claim holds on every board, not just the ones you tested.

**Two corrections to your own numbers.** Neither changes the code.

**It is nine boards a user can reach, not eleven.** Eleven is right as a count of entries in `control_boards`, and all 20 routines are really there. But `PBVA` and `PBX1` have **zero grill models pointing at them** — along with `PBL2`, they are declared boards no model in the catalogue uses:

```
declared in grills.json: 20
reachable via get_grills(): 17
declared but unused: PBL2, PBVA, PBX1
```

So the fix is still correct for all eleven, and should stay that way in case a model ever points at them — but no grill in the catalogue is affected on those two today. This is the same trap that got my `41/27` count wrong on ha-pitboss#341, in the opposite direction: raw `grills.json` over-counts where `get_grills()` tells you what is real.

**"All 20 such routines share the identical branch" is not quite true, and your next clause says so.** There are four distinct bodies, not one: three identical apart from the `FE0501` / `FE0502` / `FE0503` opcode, covering 19 routines — and `PBVA/set-temperature`, which is genuinely different because of the `cval`/`fval` table.

Worth knowing why that table exists, because it is not decoration: **it disagrees with the arithmetic conversion at the low end.** `cval[0] = 40` maps to `fval[0] = 100`, where the generic branch gives `round((40*1.8+32)/5)*5 = 105`. They agree at 45 → 115 and drift apart elsewhere. So on PBVA the vendor decided the formula was wrong and shipped a calibration; if a model is ever added on that board, that is the reason not to "simplify" it into the common form.

## What I checked beyond the diff

* The arithmetic in your new tests, independently: 100C → 212F → snapped to 210 → `FE0501020100FF`; 70C → 158F → snapped to 160 → `FE0502010600FF`, which is also exactly what `cmd(160, True)` produces. Both assertions are pinning real vendor behaviour rather than a number someone typed.
* The board each test model actually sits on — `PB1000R1` → PBV, `PB1020DX` → PBL3, `PB1600PS1` → PBL. All three match what the comments claim.
* `_is_fahrenheit()` defaults to `True` when no state has arrived, so the no-state path is passthrough — byte-for-byte today's behaviour. That is the property that makes this safe to land, and it is worth that it is a pre-existing helper rather than something introduced here.
* 739 tests, ruff, ruff format and mypy all green.

## One thing I am not able to check

I have not sent a converted setpoint to a grill in Celsius mode. The premise — that these MCUs always speak Fahrenheit — rests on the vendor's own routines converting before transmitting, which is strong evidence about what the app does and therefore what the MCU expects, but it is inference from their code rather than an observation of the hardware. If it is wrong, it is wrong in the direction of setting a Celsius grill to roughly double the intended temperature, so it is worth someone confirming on a C-mode grill after the next release rather than assuming the tests settle it.

Labelled `bug` — it was unlabelled, and this fixes wrong bytes on the wire rather than adding anything. Merging.

Issue #542, opened 2026-08-03
Transport.send_command is annotated -> dict but returns None for result-less handlers


dknowles2:
`Transport.send_command` is annotated `-> dict`, but it returns `None` whenever the firmware handler has nothing to return.

https://github.com/dknowles2/pytboss/blob/main/pytboss/transport.py#L99-L101

```python
async def send_command(
    self, method: str, params: dict, *, timeout: float | None = DEFAULT_TIMEOUT
) -> dict:
```

The reply is resolved with `payload.get("result")` (`transport.py:178`), so a handler that ends in `return null` — or a Mongoose reply that carries no `result` key at all — resolves the future to `None`.

That line came in with #534, which was the right fix: the previous `payload["result"]` raised `KeyError` on exactly those replies. But it made the return annotation wrong rather than merely optimistic, and nothing catches it — mypy trusts the declaration, and every test that exercises this path mocks `send_command` with a dict.

## How it surfaced

On #530. `request_fast_updates()` wrapped `PB.WiFiAwakeWDT`, whose handler ends in `return null`:

```js
RPC.addHandler("PB.WiFiAwakeWDT", function (params) {
  if (!checkPassword(params)) { return { error: 401, message: "Unauthorized" }; }
  wsWDT = 5 * 60;
  wsTime = 0;
  return null;
});
```

The wrapper inherited `-> dict` from the base and was corrected to `-> None` there, so #530 is fine. The base annotation is still wrong for every other caller.

Confirmed rather than reasoned:

```
request_fast_updates() returned: None | declared -> dict
```

## Which handlers are affected

`fake_firmware/init.js` has several that return nothing, and it is the repo's model of the real images rather than the images themselves — @trailro has the vendor firmware and I do not, so the real count may be larger. `PB.WiFiAwakeWDT` is the confirmed one.

## Options

1. `-> dict | None`, and let callers narrow. Honest, but pushes the question onto every call site including the many that genuinely do get a dict.
2. Keep `-> dict` for the common path and add a separate method for result-less calls. More surface for little gain.
3. `-> Any`. Truthful and useless.

I lean towards 1. The type is genuinely a union and pretending otherwise is what produced this.

Whichever way it goes, a test that drives a result-less handler through a real transport rather than a mocked `send_command` is the part that stops it regressing — mocks returning `{}` are what hid it.

Raised against ourselves per the discussion on #530; not @trailro's to fix.

trailro:
[...]

Issue #540, opened 2026-08-02
Transport lifecycle rough edges (BLE + WSS)


trailro:
A collection of smaller lifecycle/robustness issues found while reviewing the transports. None is urgent; filing them together so they don't get lost. Happy to send PRs for any of these if wanted.

### WSS: not reusable after `disconnect()`
- `disconnect()` closes the owned `ClientSession`, so a later `connect()` fails on a closed session.
- `_subscribed` is never cleared, so a reconnect via `connect()` returns before the subscribe loop is actually reading again. *(Partially addressed by #536, which clears state on connect; the owned-session part remains.)*
- Calling `connect()` twice leaks the first `_subscribe` task.

### BLE: debug-log parser fragility (`ble.py`, `_on_debug_log_received`)
- `data.decode().split()` assumes the payload contains no spaces. A `<==PBD:` vdata payload whose JSON contains a space either splits into >3 parts (silently dropped) or exactly 3 parts, in which case `int(tail[1:-1])` raises `ValueError` inside the notification callback. Current mJS firmware emits compact JSON so this mostly works today, but the parse is one space away from breaking.
- Related: the existing `# TODO: I think we want to decode this?` on the vdata path.

### BLE: connection-state hygiene (`ble.py`)
- `disconnect()` leaves `_ble_client` set; a command sent afterwards fails with a raw `BleakError` instead of `NotConnectedError`.
- `connect()` sets `_is_connected = True` before `start_notify()`; if notify registration fails, the object claims to be connected without a working receive path.
- `_on_disconnected` calls `self._loop.create_task(...)`; during interpreter/loop shutdown this can raise inside bleak's callback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

trailro:
Validated all seven against current `main` before touching anything. All still stand; two are now up as PRs.

**[#547](https://github.com/dknowles2/pytboss/pull/547) — the three BLE items.** Fixing the client-release one surfaced something this issue got wrong: `_send_prepared_command` did `if self._ble_client is None: return`, so a command with no client was **silently swallowed** and the caller waited out its full timeout. I wrote that it "fails with a raw `BleakError`"; it does not, it hangs. It raises `NotConnectedError` now.

Also worth correcting: the `create_task` in `_on_disconnected` is **mine**, from #528. There was none before it — `git show 8975a16^:pytboss/ble.py` confirms. I filed a hazard I had introduced two hours earlier.

[...]

PR #543, opened 2026-08-03
Add a local HTTP transport

dknowles2:
Adds `HttpConnection`, a transport that speaks Mongoose OS RPC to a grill directly on the local network, removing the Dansons cloud from the path.

Towards #505.

```python
from pytboss import HttpConnection, PitBoss

conn = HttpConnection("192.168.1.50")
boss = PitBoss(conn, "PBV4PS2")
await boss.start()
print(await boss.get_state())
```

## Scope: this does not turn anything on

The endpoint is gated by the firmware's `http.enable`, and **that is per-unit configuration, not a property of the model or the firmware.** @trailro's table in [ha-pitboss#344](https://github.com/dknowles2/ha-pitboss/issues/344) has two grills on the same board and the same 0.5.7 with opposite values — @dknowles2's PBV4PS2 has it off, theirs has it on. So no supported-model list could be correct, and nothing here writes config to change it.

`connect()` sends an unauthenticated `RPC.Ping` and raises `NotConnectedError` if nothing answers. Deciding whether a given grill has the endpoint is the caller's job, over a transport that already works:

```python
http = (await boss.config.get("http"))["http"]
if http["enable"]:
    ...  # an HttpConnection will work against this grill
```

That is the discovery step for a Home Assistant option later; it is deliberately not in this PR.

## Two differences from the other transports

Both are documented on the class rather than left to be found:

**There is no push channel.** HTTP is strictly request/response, so `subscribe_state()` and `subscribe_vdata()` register callbacks that never fire. `PitBoss.get_state()` issues a live RPC and updates the cache the rest of the API reads, so polling is the way to observe state and nothing else behaves differently. A test asserts the silence rather than leaving it implied.

**There is no reconnect loop**, because there is no connection to lose. An unreachable grill raises `NotConnectedError` on the next call and starts working again by itself when it returns.

## A reply is matched to its request, not to the echoed id

Over HTTP the response provably belongs to the request that produced it. So a firmware that echoes a different id — or none — cannot leave the caller waiting out a 30-second timeout for a reply already in hand. Confirmed by mutation: removing that coercion turns `test_a_reply_with_the_wrong_id_still_resolves` from a 0.04s pass into a timeout failure.

## Testing

Driven against a **real aiohttp server** standing in for `/rpc` rather than a mocked `ClientSession`, so the JSON envelope, status codes and content type are exercised as the firmware presents them — including Mongoose's habit of not setting a JSON content type.

13 tests covering: the ping on connect, round-tripping, `Unauthorized` on a password-checked method, an error payload, a non-200 status, an unreachable host, the id mismatch above, both content types, callbacks staying silent, and reuse after `disconnect()`.

That last one is worth naming. #540 records that `WebSocketConnection` cannot be reconnected after `disconnect()`, because it closes a session it owns and then reuses it. This transport rebuilds an owned session on `connect()` **and** clears it on `disconnect()` — two independent guards, so I checked the test has teeth by breaking both, at which point it fails as it should. Breaking either one alone still passes, which is the point of having two.

A result-less reply is also covered (`PB.WiFiAwakeWDT` returns null), which is #542's territory: `send_command` is annotated `-> dict` and returns `None` here. This PR does not change that annotation — it just does not pretend otherwise.

755 tests pass, `ruff check`, `ruff format --check` and `mypy .` clean.

## What is not verified

**No real grill has answered this.** @dknowles2's has `http.enable` off, and I have no access to one that has it on. The envelope shape, `/rpc` path and error format come from Mongoose's documented RPC service and from the probe script in `scripts/probe_local_rpc.py`; the decoders are the same ones the websocket transport already feeds, which is the part #505 set out to establish.

@trailro and @Giordano22 both have grills with the endpoint enabled — a single `HttpConnection(...)` + `get_state()` against real hardware would settle it, and is worth more than any test here.

### DIFF
--- a/pytboss/__init__.py
+++ b/pytboss/__init__.py
@@ -2,4 +2,5 @@
 
 from .api import PitBoss  # noqa: F401
 from .ble import BleConnection  # noqa: F401
+from .http import HttpConnection  # noqa: F401
 from .wss import WebSocketConnection  # noqa: F401
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -13,7 +13,7 @@ here turns it on: `connect()` fails against a grill where it is off, and the
 caller is expected to have established that the endpoint exists first, over a
 transport that already works. `Config.Get` over Bluetooth reports it::
 
-    http = (await boss.config.get("http"))["http"]
+    http = await boss.config.get_config("http")
     if http["enable"]:
         ...  # an HttpConnection will work against this grill
 
@@ -137,10 +137,16 @@ class HttpConnection(Transport):
                     )
                 # Mongoose does not always set a JSON content type.
                 payload: Any = await resp.json(content_type=None)
-        except ClientError as ex:
+        except (ClientError, TimeoutError) as ex:
             # The grill is unreachable rather than refusing the call. Reported
             # as a lost connection so callers treat it the way they treat a
             # dropped websocket.
+            #
+            # `TimeoutError` as well as `ClientError`: a refused connection
+            # raises the latter, but a host that simply does not answer --
+            # unplugged, asleep, or on another network -- times out instead,
+            # and `ClientTimeout` surfaces that as `asyncio.TimeoutError`,
+            # which is not a `ClientError`. Silence is the common case.
             self._connected = False
             raise NotConnectedError(f"Could not reach {self._url}: {ex}") from ex
         _LOGGER.debug("<-- %s", payload)
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -43,6 +43,7 @@ it needs an answer for every grill rather than a raise for some.
 See https://github.com/dknowles2/pytboss/issues/505.
 """
 
+import asyncio
 import logging
 from asyncio import AbstractEventLoop
 from typing import Any
@@ -103,6 +104,7 @@ class HttpConnection(Transport):
         # the grill answered the last time we spoke to it.
         self._started = False
         self._connected = False
+        self._pending: set[asyncio.Task] = set()
 
     async def connect(self) -> None:
         """Verifies the grill answers RPC at this address.
@@ -139,6 +141,9 @@ class HttpConnection(Transport):
         """
         self._started = False
         self._connected = False
+        for task in list(self._pending):
+            task.cancel()
+        self._pending.clear()
         await self._close_session()
 
     async def _close_session(self) -> None:
@@ -160,6 +165,38 @@ class HttpConnection(Transport):
             and self._session is not None
             and not self._session.closed
         )
+
+    async def send_command_without_answer(
+        self, method: str, params: dict, *, timeout: float | None = None
+    ) -> None:
+        """Sends a command without waiting for the reply.
+
+        Issued in the background, unlike the other transports, where writing
+        to the socket is the whole of it. Here a request is only finished
+        when its response has been read, and the callers that choose this
+        method choose it for commands the board cannot answer: `Sys.Reboot`
+        is gone before it could, so awaiting the exchange means waiting out
+        the full timeout and then raising `NotConnectedError` at a caller
+        that is not expecting one.
+
+        :param method: The method to call.
+        :param params: Parameters to send with the command.
+        :param timeout: Ignored. Nothing is awaited, so nothing can elapse.
+        """
+        cmd = await self._prepare_command(method, params)
+        task = self._loop.create_task(self._send_and_forget(cmd))
+        # Held so the task is not collected mid-flight, and so `disconnect()`
+        # can stop one that is still going.
+        self._pending.add(task)
+        task.add_done_callback(self._pending.discard)
+
+    async def _send_and_forget(self, cmd: dict) -> None:
+        try:
+            await self._send_prepared_command(cmd)
+        except Exception as ex:  # noqa: BLE001
+            # Nobody is waiting to be told, and for a command the board was
+            # never going to answer this is the expected ending.
+            _LOGGER.debug("No answer to %s, as expected: %s", cmd["method"], ex)
 
     async def _send_prepared_command(self, cmd: dict) -> None:
         # `_started` rather than `_connected`: a request that failed leaves
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -51,7 +51,7 @@ from typing import Any
 from aiohttp import ClientError, ClientSession, ClientTimeout
 
 from .exceptions import NotConnectedError, RPCError
-from .transport import DEFAULT_TIMEOUT, Transport
+from .transport import DEFAULT_TIMEOUT, Transport, _rpc_error
 
 _LOGGER = logging.getLogger("pytboss")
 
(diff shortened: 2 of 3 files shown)


trailro:
Ran it against a real grill — PB1600PS1, PBL board, firmware 0.5.7, `http.enable` true. **It works**, and it found two things.

## What passed

```
start()                    OK, 103 ms
is_connected()             True
PB.GetFirmwareVersion      {'firmwareVersion': '0.5.7'}
get_state()                26 fields, 320 ms
                           moduleIsOn=False  isFahrenheit=False
                           grillTemp=21  grillSetTemp=82  p1Target=10
PitBoss._state populated   True
disconnect()               OK, is_connected() -> False
connect() again            OK
get_state() x8             median 342 ms (min 322, max 603)
```

The reconnect-after-`disconnect()` path you built two guards for and could not verify does work on hardware. So does the cache update, which matters because it is the whole reason a polling-only transport is usable at all.

Two of your tests I could also confirm against the grill rather than a stand-in server:

- **Callbacks really do stay silent.** Three `get_state()` calls with a state subscriber registered fired it zero times.
- **Error payloads come back as you handle them.** `No.SuchMethod` gives `RPCError` with `code=404` and `"No handler for No.SuchMethod"` — the code populated, since #524.

Latency is in the same range as the websocket path on the same grill, so nothing is lost by the request/response shape.

## Two defects

**1. `TimeoutError` escapes `connect()`.** The docstring promises `NotConnectedError` if nothing answers, and that holds for a refused connection but not for a silent one:

```
192.168.1.253 (silent, nothing listening)   -> TimeoutError          isinstance NotConnectedError: False
127.0.0.1:9   (connection refused, RST)     -> NotConnectedError     ✓
```

`ClientTimeout` surfaces as `asyncio.TimeoutError`, which is not a `ClientError`, so the `except ClientError` in `_send_prepared_command` does not see it. The silent case is the common one — a grill that is unplugged, asleep or on a different VLAN does not send an RST, it just does not answer. As it stands `connect()` also leaves `_connected` True on that path, because the reset happens in the same handler.

Catching `TimeoutError` alongside `ClientError` would cover it.

**2. The discovery snippet in the docstring does not run.** Two problems in one line:

```python
http = (await boss.config.get("http"))["http"]   # as written
```

`Config` has no `get` — the method is `get_config`. And `get_config("http")` returns the subtree itself, not a dict keyed by the section:

```python
>>> await boss.config.get_config("http")
{"enable": true, "listen_addr": "80", "document_root": "/", ...}
```

So it is:

```python
http = await boss.config.get_config("http")
if http["enable"]:
    ...
```

Worth getting right precisely because that snippet is the gate — it is what a Home Assistant config flow would copy, and it is the only thing standing between "offer the local option" and offering it to grills where it cannot work.

## Everything else checks out

The per-unit framing is right and matches what @dknowles2 and I found. Not writing config to enable the endpoint is the right call, and I would not want it behind a checkbox either.

The id coercion is sound here — I could not force a mismatch on this firmware, so your mutation test remains the evidence for it, but the reasoning holds: over HTTP the reply provably belongs to the request.

One observation rather than a request: `p1Target` came back as `10` with the grill off and no probe plugged in. That is the board reporting something stale rather than anything this transport does — it reads identically over the other transports — but if anyone builds on `pNTarget` while the grill is cold, that is the value they will get.

Happy to re-run anything else against the grill; it is the one thing I can contribute here that the tests cannot.

trailro:
Both fixes from my review are up as [#546](https://github.com/dknowles2/pytboss/pull/546), targeting `http-transport` rather than `main` — so you can merge it into this PR with one click, or take the diff and write them yourself, or ignore it. I cannot push to your branch, so a PR against it was the only way to offer code rather than a description.

Nothing else in #543 is touched. The id coercion, session ownership and silent-callbacks behaviour all stand as written, and all three check out on hardware.

Verified the fixes on the same grill, not just against the test server: a silent host now raises `NotConnectedError` instead of `TimeoutError`, the docstring snippet executes as written and returns `True`, and `get_state()` still comes back with 26 fields afterwards.

trailro:
One more thing from the hardware side, and it turns out to be about a grill I do **not** have.

I pulled the other firmware line and went through it, because `http.enable` framing assumes every grill runs Mongoose. Not all of them do.

## There are two firmware families, and only one of them can ever serve this

Dansons ships a second, unrelated firmware: not Mongoose, not mJS, but **C on ESP-IDF v4.4.3/v4.4.4**, versioned `16.x` instead of `0.x`. The images are public and listable at `s3://dansons-mobile/firmware/` — 13 of them, boards PBC2/PBD/PBE/PBL2/PBT, 16.2.0 through 16.8.3. I archived the set and read them.

**Those grills have no local RPC endpoint at all**, and no setting can give them one:

* `esp_http_server` **is not linked in**. The `ESP_ERR_HTTPD_*` strings are present, but only as `esp_err_to_name` table entries — the component's own source paths (`components/esp_http_server/...`) are absent, where the websocket client's (`esp_transport_ws`) are there. I nearly reported the opposite off a plain `strings` grep.
* WebSocket is **client-only** (`WEBSOCKET_CLIENT`), used for one outbound connection to `wss://socket.dansonscorp.com/from/`.
* No mDNS, no `bind`/`listen` in application code, no `/rpc` path string anywhere.

So an ESP-IDF grill is reachable over Bluetooth or the cloud relay, and by nothing else.

## Which makes the rule two-level rather than one

Your framing is right *within* the Mongoose family — @dknowles2's PBV4PS2 and my PB1600PS1 really are the same board and firmware with opposite `http.enable` — but the catalogue-level rule is:

1. **Firmware family.** ESP-IDF grills: never, and no configuration changes that.
2. **Per-unit `http.enable`,** within the Mongoose family only.

## The practical consequence is in the discovery snippet

```python
http = await boss.config.get_config("http")
if http["enable"]:
```

That assumes Mongoose's config schema. The ESP-IDF firmware does have `Config.Get`, but its schema is only `wifi.sta.{ssid,pass}`, `app` and `version` — **there is no `http` section to read.** (The one bare `http` string in those images belongs to `esp_http_client`'s scheme table, next to `https` and `WWW-Authenticate`.) So on an IDF grill that snippet does not return `{"enable": false}`; it fails, and the shape of the failure depends on how their `Config.Get` handles an unknown key, which I cannot test without one of those grills.

Since the snippet is what a config flow copies, "offer the local option or not" would end up being decided by an exception rather than by an answer. Same reasoning as the last fix: it is documentation that gates a feature, so it is code.

PR against `http-transport` shortly, same as #546 — docstring only, no behaviour change. `connect()` already does the right thing against these grills: nothing listens on port 80, the connection is refused, and it raises `NotConnectedError` exactly as intended.

One caveat on my own evidence: this is static analysis of stripped release binaries, and absence of a linked component is a stronger claim than absence of a string. I checked it the way described rather than by asking one of those grills, because I do not have one. @Giordano22 or anyone with a PBD/PBT/PBE/PBC2/PBL2 could settle it in one request — if `http://<ip>/rpc` answers on one of those, I am wrong.

dknowles2:
Branch state, since this has accumulated a few things while waiting.

**It now carries two of @trailro's PRs against it**, both merged into `http-transport` rather than `main`:

* **#546** — two bugs of mine, found by running this against a real PB1600PS1. `except ClientError` missed `asyncio.TimeoutError`, so a *silent* grill (unplugged or asleep, which sends no reset) raised a bare `TimeoutError` and left `_connected` True — a coordinator would have kept reporting the grill available while every call timed out. And the discovery snippet in the module docstring called a `Config.get` that does not exist, then indexed a subtree that is already the subtree.
* **#558** — the availability rule is two-stage, not one. Beyond `http.enable`, there is a second Dansons firmware line (C on ESP-IDF, versioned `16.x`) on PBC2, PBD, PBE, PBL2 and PBT that links no HTTP server at all. That is **17 reachable models plus the two PBL2-board grills**, about an eighth of the catalogue, which cannot serve this endpoint under any configuration. My "per-unit, no model predicts it" framing was true of the Mongoose family and stated as the whole rule.

I have merged `main` in as well, which matters for one specific reason: #551 narrowed `Config.get_config` to return `{}` rather than `None` for a result-less reply, and the documented snippet depends on that. Before the merge it raised on a `None` reply; now it answers for every shape:

```
reply={'enable': True}   -> True
reply={'enable': False}  -> False
reply={}                 -> False
reply=None               -> False
raises RPCError          -> False
```

**782 tests, ruff, `ruff format --check` and mypy clean.**

The snippet is also now executed by a test rather than only rendered, which is the direct lesson from it having been wrong: documentation that decides whether a feature is offered to a user is code.

Still unverified, and worth carrying into any release note: **nobody has confirmed the ESP-IDF claim against hardware.** It is static analysis of stripped binaries — a stronger form of evidence than a missing string, but not a device saying no. One request to `http://<ip>/rpc` from anyone with a PBD, PBT, PBE, PBC2 or PBL2-board grill settles it either way.

PR #546, opened 2026-08-03
Treat a silent grill as unreachable, and fix the discovery snippet

trailro:
The two fixes from my review of #543, as a PR against that branch rather than a comment — merge it in if you want it, ignore it if you would rather write them yourself. I cannot push to `http-transport` directly, so this is the only shape I can offer them in.

Both were found by running #543 against a real grill (PB1600PS1, PBL, 0.5.7, `http.enable` on), and both fixes are verified against the same grill rather than only the test server.

**1. A silent grill now reports `NotConnectedError`.**

`ClientTimeout` surfaces as `asyncio.TimeoutError`, which is not a `ClientError`, so the existing handler did not see it:

```
192.168.1.253 (silent, nothing listening)  before: TimeoutError        after: NotConnectedError
127.0.0.1:9   (refused, RST)               before: NotConnectedError   after: unchanged
```

The silence case is the common one — a grill that is unplugged, asleep or on another network does not send a reset — and it also left `_connected` True, because the reset lives in the same handler.

Test is in your style, a real server whose handler never answers, so it exercises the aiohttp timeout path rather than a mocked raise. Verified it fails without the fix.

**2. The discovery snippet in the module docstring now runs.**

```diff
-    http = (await boss.config.get("http"))["http"]
+    http = await boss.config.get_config("http")
     if http["enable"]:
```

`Config` has no `get`, and `get_config("http")` returns the subtree itself rather than a dict keyed by section. Executed against the grill exactly as written now:

```
>>> http = await boss.config.get_config("http")
>>> http["enable"]
True
```

Worth being exact about because that snippet is the gate — it is what a Home Assistant config flow copies, and it decides whether the local option is offered to a grill where it cannot work.

**Nothing else changed.** I did not touch the id coercion, the session ownership, the callbacks-stay-silent behaviour or any of your tests; those all check out, including on hardware.

756 tests, ruff, `ruff format --check` and mypy clean. Re-ran the grill afterwards: `get_state()` still returns 26 fields.

### DIFF
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -13,7 +13,7 @@ here turns it on: `connect()` fails against a grill where it is off, and the
 caller is expected to have established that the endpoint exists first, over a
 transport that already works. `Config.Get` over Bluetooth reports it::
 
-    http = (await boss.config.get("http"))["http"]
+    http = await boss.config.get_config("http")
     if http["enable"]:
         ...  # an HttpConnection will work against this grill
 
@@ -137,10 +137,16 @@ class HttpConnection(Transport):
                     )
                 # Mongoose does not always set a JSON content type.
                 payload: Any = await resp.json(content_type=None)
-        except ClientError as ex:
+        except (ClientError, TimeoutError) as ex:
             # The grill is unreachable rather than refusing the call. Reported
             # as a lost connection so callers treat it the way they treat a
             # dropped websocket.
+            #
+            # `TimeoutError` as well as `ClientError`: a refused connection
+            # raises the latter, but a host that simply does not answer --
+            # unplugged, asleep, or on another network -- times out instead,
+            # and `ClientTimeout` surfaces that as `asyncio.TimeoutError`,
+            # which is not a `ClientError`. Silence is the common case.
             self._connected = False
             raise NotConnectedError(f"Could not reach {self._url}: {ex}") from ex
         _LOGGER.debug("<-- %s", payload)
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -232,3 +232,30 @@ async def test_state_callbacks_never_fire(host):
         assert seen == []
     finally:
         await conn.disconnect()
+
+
+async def test_a_silent_grill_is_reported_as_not_connected():
+    """Unreachable and unresponsive are the same thing to a caller.
+
+    A refused connection raises `ClientError`, but a host that answers
+    nothing times out instead, and `ClientTimeout` surfaces that as
+    `asyncio.TimeoutError` -- which is not a `ClientError`. Silence is the
+    common case: a grill that is unplugged, asleep or on another network does
+    not send a reset.
+    """
+
+    async def never_answers(request: web.Request) -> web.Response:
+        await asyncio.sleep(30)
+        raise AssertionError("unreachable")
+
+    app = web.Application()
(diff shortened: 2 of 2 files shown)


trailro:
Note on the empty check list: `build-and-test.yml` triggers on `pull_request: branches: ["main"]`, so nothing runs for a PR targeting a feature branch. That is not a failure, it is CI never being asked.

So the green I am claiming is local, at the CI-pinned versions:

```
756 passed
ruff check          all checks passed
ruff format --check 33 files already formatted
mypy .              no issues
```

It will get a real run the moment this is merged into `http-transport` and #543 rebuilds. Flagging it so an empty status is not mistaken for a passing one — I would rather say CI has not seen this than let the absence read as approval.

dknowles2 (approved):
Both of these are my bugs, both are real, and you found them the only way they could have been found. Thank you for running it against an actual grill — that was the open question on #543 and I had no way to close it myself.

**The timeout handler is the worse of the two and I got it exactly backwards.** I wrote `except ClientError` and a comment claiming it covered "the grill is unreachable", having reasoned about a refused connection and not about silence. Silence is the common case: an unplugged or sleeping grill sends no reset, `ClientTimeout` surfaces as `asyncio.TimeoutError`, and that is not a `ClientError`. So the transport raised a bare `TimeoutError` at callers and — worse — left `_connected` True, because the flag is cleared in the handler that never ran. A Home Assistant coordinator would have kept reporting the grill available while every call timed out.

I reproduced it rather than taking it on the description: reverting your line to my original `except ClientError` fails your new test.

```
FAILED tests/test_http.py::test_a_silent_grill_is_reported_as_not_connected
1 failed, 13 passed
```

And the test earns it — a real server whose handler never answers, driving aiohttp's own timeout path, rather than a mocked raise that would have proved only that I can spell `TimeoutError`.

**The docstring snippet was wrong twice over**, which is worse than it sounds because that snippet is the whole gate. `Config` has no `get` — the method is `get_config` — so it would have raised `AttributeError` on the first line. And `get_config("http")` returns the subtree itself, so my `["http"]` indexing was wrong on top of that. Confirmed against the repo:

```
Config methods: ['get_config', 'get_info', 'save_config', 'set', 'set_wifi_credentials', ...]
```

I wrote that snippet to be copied into a Home Assistant config flow and never executed it. Documentation that decides whether a feature is offered to a user is code, and I did not treat it as code.

## Verified here

* 756 tests, `ruff check`, `ruff format --check` and `mypy .` all clean on this branch.
* CI has not run and will not: `build-and-test.yml` triggers only on pull requests targeting `main`, and this targets `http-transport`. Unlike ha-pitboss, the pytboss suite runs locally, so the numbers above are a real check rather than a stand-in for one. #543's own CI will run once this lands on the branch.
* Nothing else touched — id coercion, session ownership and the silent-callbacks behaviour are unchanged, as you say.

**The more valuable thing in this PR is the line I almost skipped over:** `get_state()` returning 26 fields from a real PB1600PS1 on 0.5.7. That is the first evidence that any of #543 works against hardware. Everything I could offer was a test server built from Mongoose's documentation and my own reading — which is exactly the kind of thing that agrees with itself and not with the world. It matters that the decoders came back with a full frame unchanged.

Labelled `bug`. Merging into `http-transport` so #543 carries both fixes; whether #543 itself lands is David's call.

PR #547, opened 2026-08-03
Fix BLE connection-state hygiene

trailro:
First of two from #540. Three connection-state problems in `ble.py`, all validated against current `main` before writing anything.

**1. `connect()` reported success before there was a receive path.** `_is_connected = True` was set before `start_notify()`. If notify registration failed, the transport claimed to be connected with nothing listening — every command would be sent and never answered, while `is_connected()` said the link was fine. Moved after both registrations.

This is the one with real consequence: on Bluetooth a coordinator would report the grill online and simply fail every call.

**2. `disconnect()` kept the client.** `_ble_client` stayed set, so a later command reached a disconnected bleak client. Released now.

**3. `_on_disconnected` scheduled work unconditionally.** Bleak calls it synchronously, including while the loop is going away, where `create_task` raises inside bleak's own callback. Guarded.

**That third one is mine.** There was no `create_task` there before [#528](https://github.com/dknowles2/pytboss/pull/528) — I added it to fail in-flight commands, and in doing so introduced the shutdown hazard I then filed against myself. Confirmed with `git show 8975a16^:pytboss/ble.py`.

### One thing #540 did not know about, and it is worse than what it described

Fixing (2) surfaced this. `_send_prepared_command` did:

```python
if self._ble_client is None:
    return
```

So a command with no client was **silently swallowed**: never put on the wire, and the caller then awaited a reply that could not come until its timeout expired. #540 said a post-disconnect command "fails with a raw `BleakError`" — it does not, it hangs. It now raises `NotConnectedError`, which is what every other transport does.

**This is a behaviour change and an existing test encoded the old one.** `test_send_prepared_command_no_client` asserted the silent return; it now asserts the raise. Flagging it rather than letting it look like a test that needed adjusting.

### Tests

Four, each verified to fail without its fix: a failed notify does not report connected; a command after disconnect raises `NotConnectedError`; a disconnect notification during loop shutdown does not raise; and the updated no-client case.

745 tests, ruff, `ruff format --check` and mypy clean. Merges cleanly with #543 and with the WSS half of #540, which I have sent separately.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -96,9 +96,14 @@ class BleConnection(Transport):
         # Guarded because this also fires while the loop is shutting down,
         # where scheduling raises inside bleak's own callback.
         if not self._loop.is_closed():
+            # The coroutine is built before `create_task` is called, so it has
+            # to be closed explicitly when scheduling fails -- otherwise it is
+            # abandoned unawaited and warns from wherever the collector runs.
+            coro = self._fail_pending_commands()
             try:
-                self._loop.create_task(self._fail_pending_commands())
+                self._loop.create_task(coro)
             except RuntimeError:  # pragma: no cover - loop shutting down
+                coro.close()
                 _LOGGER.debug("Loop is closing; not failing pending commands.")
         if not self._reconnecting and self._disconnect_callback is not None:
             self._disconnect_callback(client)
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -17,6 +17,7 @@ import bleak_retry_connector
 from bleak import BleakClient, BleakGATTCharacteristic, BLEDevice
 from bleak_retry_connector import BleakClientWithServiceCache
 
+from .exceptions import NotConnectedError
 from .transport import Transport
 
 _LOGGER = logging.getLogger("pytboss")
@@ -80,16 +81,25 @@ class BleConnection(Transport):
             name=self._ble_device.name or "<unknown>",
             disconnected_callback=self._on_disconnected,
         )
-        self._is_connected = True
+        # Only once the notifications are registered: until they are there is
+        # no receive path, so a command would be sent and never answered while
+        # `is_connected()` claimed otherwise.
         await self._ble_client.start_notify(CHAR_RPC_RX_CTL, self._on_rpc_data_received)
         await self._ble_client.start_notify(CHAR_DEBUG_LOG, self._on_debug_log_received)
+        self._is_connected = True
 
     def _on_disconnected(self, client: BleakClient) -> None:
         """Called when our Bluetooth client is disconnected."""
         _LOGGER.debug("Bluetooth disconnected.")
         self._is_connected = False
         # Bleak calls this synchronously, so the cleanup has to be scheduled.
-        self._loop.create_task(self._fail_pending_commands())
+        # Guarded because this also fires while the loop is shutting down,
+        # where scheduling raises inside bleak's own callback.
+        if not self._loop.is_closed():
+            try:
+                self._loop.create_task(self._fail_pending_commands())
+            except RuntimeError:  # pragma: no cover - loop shutting down
+                _LOGGER.debug("Loop is closing; not failing pending commands.")
         if not self._reconnecting and self._disconnect_callback is not None:
             self._disconnect_callback(client)
 
@@ -102,6 +112,10 @@ class BleConnection(Transport):
             except Exception as ex:  # noqa: BLE001
                 # Bluetooth is awful. Sometimes even disconnects fail.
                 _LOGGER.debug("Failed to disconnect: %s", ex)
+            # Released either way: a client kept after a disconnect answers a
+            # later command with a raw `BleakError` instead of the
+            # `NotConnectedError` every other transport raises.
+            self._ble_client = None
         self._is_connected = False
 
     async def reset_device(self, ble_device: BLEDevice):
@@ -125,7 +139,10 @@ class BleConnection(Transport):
 
     async def _send_prepared_command(self, cmd: dict):
         if self._ble_client is None:
-            return
+            # Raised rather than returned: swallowing the send left the caller
+            # awaiting a reply for a command that was never put on the wire,
+            # so it waited out its whole timeout to learn nothing.
+            raise NotConnectedError("Not connected")
         payload = json.dumps(cmd)
         async with self._lock:
             await self._ble_client.write_gatt_char(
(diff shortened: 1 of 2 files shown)


dknowles2 (commented):
All three fixes are real and I mutated each one independently rather than trusting the list — every fix has exactly the test that pins it:

```
revert the _is_connected move  -> test_a_failed_notify_does_not_report_connected
revert raise -> silent return  -> test_a_command_after_disconnect_reports_not_connected + test_send_prepared_command_no_client
revert the client release      -> test_a_command_after_disconnect_reports_not_connected
```

**Fix 1 is the one with teeth.** Claiming connected before `start_notify` means a failed registration leaves a transport that reports healthy and answers nothing — on Bluetooth that is a coordinator reporting the grill online while every call times out. Moving the flag after both registrations is right.

**Filing the third one against yourself and then saying so here is worth noting.** You added that `create_task` in #528 to fail in-flight commands, and it brought the shutdown hazard with it. Citing `git show 8975a16^:pytboss/ble.py` to establish it was not there before is the right way to settle that kind of question.

## One correction to the write-up, and it cuts in your favour

> #540 said a post-disconnect command "fails with a raw `BleakError`" — it does not, it hangs.

**#540 was right about `main`.** On `main`, `disconnect()` keeps `_ble_client` — I checked `git show main:pytboss/ble.py` — so a later command reaches a disconnected bleak client and does raise `BleakError`. The silent-swallow path needs `_ble_client is None`, which on `main` only happens before the first `connect()`.

The hang becomes reachable **because of your fix 2**. So the sequencing in your own summary is the accurate one — "fixing (2) surfaced this" — and the sentence after it overstates the before-state. Nothing about the change is wrong; the two fixes genuinely have to land together, which is the point worth keeping.

## One thing I would fix before merge

**This introduces a `RuntimeWarning` into a previously clean suite.**

```
tests/test_config.py::test_get_info
  RuntimeWarning: coroutine 'Transport._fail_pending_commands' was never awaited
```

`main` has none; this branch has one. It is attributed to `test_config.py` only because that is where the GC happened to run — the source is `test_a_disconnect_during_shutdown_does_not_raise`, which patches `create_task` to raise. The coroutine is constructed as the argument *before* `create_task` is called, so when that raises, the object is abandoned unawaited.

It is harmless at runtime — the loop is dying anyway — but it is a leak in exactly the path this PR adds, and it makes the suite noisy for everyone after. Two lines:

```python
coro = self._fail_pending_commands()
try:
    self._loop.create_task(coro)
except RuntimeError:  # pragma: no cover - loop shutting down
    coro.close()
    _LOGGER.debug("Loop is closing; not failing pending commands.")
```

`coro.close()` is the explicit "I am discarding this" that stops the warning.

Worth checking whether the `is_closed()` guard is still earning its place once that is in, incidentally — with the `RuntimeError` handler doing the work, the pre-check mostly covers a loop that is closed but not yet raising. No strong view; I would keep both.

## Otherwise

745 tests, ruff, `ruff format --check` and mypy clean. Flagging the changed assertion in `test_send_prepared_command_no_client` rather than letting an edited test pass unremarked is the right instinct — an existing test that encoded the old behaviour is exactly the thing a reviewer should be told about rather than left to notice.

Labelled `bug`. Happy to merge once the warning is gone; say if you would rather land it and follow up.

trailro:
Fixed at `f21fe60`, and **you are right about the write-up — #540 was correct and my correction to it was wrong.**

I checked `main` rather than take it: `disconnect()` leaves `_ble_client` set, so `_send_prepared_command` passes its `is None` check and reaches `write_gatt_char` on a disconnected client. `BleakError` is exactly what happens. The silent swallow needs `_ble_client is None`, which on `main` only holds before the first `connect()` — a state nobody sends commands from.

So the hang is something **my fix 2 creates**, not something that was already there. I over-corrected: I found the swallow while testing my own change, assumed it had always been reachable, and told you the issue had it wrong. The sequencing in the summary — "fixing (2) surfaced this" — was accurate, and then I contradicted it a paragraph later.

The practical consequence is the one you draw: fixes 2 and 3 have to land together, because releasing the client without raising would turn a wrong exception into a silent timeout. Worth stating plainly in case anyone ever splits them.

**The `RuntimeWarning` is gone.** Your diagnosis was exact — the coroutine is constructed as the argument, so `create_task` raising abandons an object nobody ever awaits, and the collector reports it from wherever it happens to run. `coro.close()` in the handler:

```python
coro = self._fail_pending_commands()
try:
    self._loop.create_task(coro)
except RuntimeError:  # pragma: no cover - loop shutting down
    coro.close()
```

Verified by removing `coro.close()` again: the warning comes back, and with it in place the suite is clean. `main` has none and neither does this branch now.

**On keeping the `is_closed()` pre-check:** agreed, keep both. They cover different states — a loop already closed, and one closing while the call is in flight — and this runs inside bleak's callback where an escaping exception is not ours to see. Belt and braces is the right posture in a synchronous callback we do not own.

745 tests, ruff, `ruff format --check` and mypy clean, and the suite has no warnings it did not have before.

dknowles2 (approved):
Fixed, and verified both directions rather than one:

```
with coro.close():     745 passed, 'never awaited' warnings = 0
without coro.close():  'never awaited' warnings = 1
```

So the suite is back to `main`'s cleanliness, and `coro.close()` is demonstrably what does it rather than something else having shifted.

**On the write-up correction — you went and checked `main` rather than take my word, which is the right response even when the reviewer turns out to be right.** And the conclusion you draw from it is the one that matters: the hang is something fix 2 *creates*, so fixes 2 and 3 have to land together. Releasing the client without also raising would convert a wrong-but-loud `BleakError` into a silent timeout, which is strictly worse. That is worth having in the commit history in case anyone ever splits them.

I had the three fixes mutated independently before that exchange, and each still has exactly the test that pins it:

```
revert the _is_connected move  -> test_a_failed_notify_does_not_report_connected
revert raise -> silent return  -> test_a_command_after_disconnect_reports_not_connected
revert the client release      -> test_a_command_after_disconnect_reports_not_connected
```

Agreed on keeping both guards in `_on_disconnected`. They cover different states, and this runs inside bleak's synchronous callback where an escaping exception is not ours to observe — belt and braces is the right posture in code someone else calls.

745 tests, ruff, `ruff format --check` and mypy clean, CI green, no warnings the branch did not start with. Merging.

That closes both halves of #540. Worth noting what the pair actually did: the reported symptoms were mostly the visible edge of two transports that could not be reconnected and one that lied about being connected — and between them, `ble.py` and `wss.py` now behave the way `http.py` was written to from the start, which is a better outcome than fixing the four bullet points as listed.

PR #549, opened 2026-08-03
Parse the debug log without splitting the payload

trailro:
The last item in #540, and **I was wrong to say it should be left.**

I argued it was speculative hardening I could not test without synthesising a payload the firmware does not emit. That was wrong on both counts: the payload is reachable through the library's own API, and the parser is a pure function of its input, so testing it with an input a consumer can cause is entirely fair. I talked myself out of it on a bad premise.

**It is reachable through `set_virtual_data`.** The grill echoes back what was written, so storing any string with a space in it produces a debug-log line the parser cannot handle:

```
>>> await conn._on_debug_log_received(None, bytearray(b'<==PBD:  {"note":"hello world"}'))
ValueError: invalid literal for int() with base 10: 'orld"'
```

Raised inside the bleak notification callback. Split on every space, `{"note":"hello world"}` becomes two parts, the second is read as the length field, and `int()` gets `orld"`.

The other shape is quieter and worse: a payload with two spaces gives four parts, falls through to the `else`, and is dropped with no log line at all.

**The fix stops splitting the payload.** Head comes off at the first space, an optional trailing `[len]` comes off the end, and everything between is the payload — spaces and all.

Two tests, both verified to fail against `main`:

- a JSON string value with a space arrives intact
- the checksum measures the whole payload rather than the run up to the first space

Note the length marker is `[5]`, square brackets. I assumed parentheses first and the existing `test_debug_log_bad_checksum` caught it immediately — the old `tail[1:-1]` worked for either, which is why nothing in the code says which it is.

**The `# TODO: I think we want to decode this?` on the vdata path is untouched.** It is a separate question — whether the callback should receive parsed JSON rather than a string — and it changes a public callback's contract, so it did not belong here.

751 tests, ruff, `ruff format --check` and mypy clean. With this, #540 has nothing left in it.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -175,23 +175,29 @@ class BleConnection(Transport):
         self, unused_char: BleakGATTCharacteristic, data: bytearray
     ):
         _LOGGER.debug("Debug log received: %s", data)
-        parts = data.decode("utf-8").split()
-        if len(parts) == 3:
-            head, payload, tail = parts
-            checksum = int(tail[1 : len(tail) - 1])
-            if len(payload) != checksum:
-                # Bad payload; ignore.
-                _LOGGER.debug(
-                    "Ignoring message with bad checksum (%d != %d)",
-                    len(payload),
-                    checksum,
-                )
-                return
-        elif len(parts) == 2:
-            head, payload = parts
-        else:
+        # Split off the head and an optional trailing "[len]" rather than on
+        # every space: a virtual data payload is JSON, and any string value in
+        # it with a space in it would otherwise either be dropped or parsed as
+        # the length. Both are reachable through `set_virtual_data`, since the
+        # grill echoes back what was written.
+        head, _, rest = data.decode("utf-8").strip().partition(" ")
+        payload = rest.strip()
+        if not payload:
             # Unknown payload; ignore.
             return
+        if payload.endswith("]"):
+            body, _, tail = payload.rpartition("[")
+            if body and tail[:-1].isdigit():
+                payload = body.strip()
+                checksum = int(tail[:-1])
+                if len(payload) != checksum:
+                    # Bad payload; ignore.
+                    _LOGGER.debug(
+                        "Ignoring message with bad checksum (%d != %d)",
+                        len(payload),
+                        checksum,
+                    )
+                    return
         if head == "<==PB:" and self._state_callback:
             status_payload = temperatures_payload = None
             match payload[:4]:
--- a/tests/test_ble.py
+++ b/tests/test_ble.py
@@ -434,3 +434,47 @@ async def test_disconnect_notification_during_shutdown_does_not_raise(
         conn._on_disconnected(mock_bleak_client)
 
     assert not conn.is_connected()
+
+
+@mock.patch("bleak_retry_connector.establish_connection")
+@mock.patch("bleak.BleakClient", spec=True)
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
**Reproduced both failure shapes against `main` before reading the fix**, and they are exactly as you describe — including the quieter one, which is the worse of the two:

```
'<==PBD:  {"note":"hello world"}'  -> ValueError: invalid literal for int() with base 10: 'orld"'
'<==PBD:  {"a":"x y z"}'           -> no error, silently dropped
```

One space raises inside bleak's notification callback; two spaces fall through the `else` and vanish with no log line at all. On the branch both arrive intact.

**And the reachability argument holds.** `fake_firmware/init.js:173` is `print("<==PBD: ", JSON.stringify(data))`, so the line carries whatever was stored — and `set_virtual_data` lets a consumer store any string. That is not synthesising a payload the firmware does not emit; it is emitting one through the library's own front door.

**Retracting your own "leave it" is the right call and the reasoning behind the retraction is the useful part.** "I cannot test it without an unrealistic input" and "the input is reachable through our own API" are very different claims, and you had merged them. A parser is a pure function of its input, so the bar is whether a consumer can cause the input, not whether you have seen the device produce it.

## I probed the new parser for false positives

The risk in `rpartition("[")` is a payload that legitimately ends in `]`. Every shape I could construct behaves:

```
'<==PBD:  [1,2,3]'              -> '[1,2,3]'        array payload survives (empty body -> no checksum branch)
'<==PBD:  {"a":"b"} [9]'        -> '{"a":"b"}'      checksum honoured
'<==PBD:  {"a":"x y"} [11]'     -> '{"a":"x y"}'    length counts the space
'<==PBD:  {"a":"x y"} [9]'      -> dropped          wrong length still rejected
'<==PBD:  {"k":"v[2]"}'         -> '{"k":"v[2]"}'   brackets inside a string are safe
```

Nested arrays are safe too — `[[5]]` and `[1,[2]]` both leave a non-digit tail after the last `[`, so the branch is skipped. The `body and tail[:-1].isdigit()` pair is doing real work; either half alone would misfire.

## One thing neither of us can settle from this repo

**`fake_firmware` emits no length marker at all** — `print("<==PBD: ", ...)` and `print("<==PB: ", status)` produce head and payload, nothing else. So the `[len]` form is attested here only by `test_debug_log_bad_checksum`, which predates both of us, and nothing in the tree says whether the firmware's length counts a payload's internal spaces.

Your reading is the natural one and your test pins it — but nobody has seen a real payload that has both a space *and* a checksum, because until now a payload with a space could not survive the parser to be checked. So that assertion is a reasonable inference rather than an observation, and if a real capture ever contradicts it, `test_debug_log_checksum_counts_the_whole_payload` is the line to revisit. Worth writing down given you have the vendor images and the hardware and I have neither.

Good that the parentheses-versus-brackets slip was caught by an existing test rather than by review. `tail[1:-1]` working for either is exactly the kind of thing that hides an assumption — the new code names it.

Leaving the `# TODO: I think we want to decode this?` alone was right: changing what a public callback receives is a separate decision from parsing the line correctly.

751 tests, ruff, `ruff format --check` and mypy clean; CI green. Labelled `bug`. Merging.

**That is #540 empty.** Between #547, #548 and this, the three transports now agree on connection state, reuse and payload handling — which is more than the issue asked for.

PR #550, opened 2026-08-03
Add turn_grill_on()

trailro:
For [ha-pitboss#322](https://github.com/dknowles2/ha-pitboss/issues/322), where you said this should have an explicit pytboss method before anything downstream uses it. Agreed then and this is that method; the Home Assistant service stays where it is until this lands in a release.

`turn_grill_on()` sends `FE0101FF`.

**It cannot mirror `turn_grill_off()`, and the reason is worth recording.** That one sends the `turn-off` slug. There is no `turn-on` slug to send — I checked the whole catalogue rather than the two boards I have:

```
turn-off       : 137 of 137 models
turn-on        :   0
turn-light-on  : 132
turn-light-off : 132
```

Every board can be told to switch off. Not one declares a way to be told to switch on. That looks like a decision rather than an oversight, which is part of why the caller should be the one deciding whether to use this.

**The byte layout leaves no ambiguity about what the command is**, independently of the third-party driver quoted in the issue:

```
turn-off        FE 01 02 FF
turn-light-on   FE 02 01 FF
turn-light-off  FE 02 02 FF
this            FE 01 01 FF
```

Third byte is the subsystem — 01 power, 02 light — fourth is on or off. This is the missing member of a pair the board already implements three quarters of. I generated those three from `grills.json` rather than reading them off a datasheet.

**Confirmed working on hardware twice**: you tested it on your PBV4PS2 and said so on the issue, and I ran it on a PB1600PS1 with the lid open and the pot clear. Neither of us is guessing about whether the board accepts it.

**No safety logic here, deliberately.** The docstring says plainly that this starts a fire in an appliance nobody may be standing next to, and that the board accepts it in any state. Deciding whether starting is a good idea belongs to the caller — in the integration that is an options toggle plus refusals on no-pellets, high-temp, ignition, fan, auger and igniter errors, on being already on, and on being disconnected. None of that belongs in a transport library.

Two tests: the hex is asserted directly rather than the decoded command, since `FE0101FF` against `FE0102FF` is the whole identity of it; and one that pins the "no board declares `turn-on`" claim against the real catalogue, so the docstring's reasoning fails loudly if a definitions refresh ever changes it.

754 tests, ruff, `ruff format --check` and mypy clean.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -437,20 +437,29 @@ class PitBoss:
 
         Sent as a raw MCU command because no board declares a slug for it. All
         137 models declare `turn-off`; **not one declares a `turn-on`**, which
-        looks deliberate on the vendor's part rather than an oversight. The
-        byte layout leaves no doubt about which command it is:
+        looks deliberate on the vendor's part rather than an oversight.
 
-        ==================  ==========
-        ``turn-off``        FE0102FF
-        ``turn-light-on``   FE0201FF
-        ``turn-light-off``  FE0202FF
-        this                FE0101FF
-        ==================  ==========
+        What the catalogue actually contains:
 
-        The third byte is the subsystem -- 01 power, 02 light -- and the
-        fourth is on or off, so this is the missing member of a pair the board
-        already implements three quarters of. Confirmed working on a
-        PB1600PS1 and a PBV4PS2.
+        ==================  ==========  ===========
+        slug                command     models
+        ==================  ==========  ===========
+        ``turn-off``        FE0102FF    137
+        ``turn-light-on``   FE0201FF    132
+        ``turn-light-off``  FE0200FF    121
+        ``turn-light-off``  FE0202FF     11
+        this                FE0101FF      0
+        ==================  ==========  ===========
+
+        The third byte is the subsystem -- 01 power, 02 light. The fourth is
+        `01` wherever a command means "on"; "off" is spelled `02` or `00`
+        depending on the board, so it is not a uniform flag. What holds is
+        the narrower claim: `01` is the only value that ever means on, and
+        `FE0101FF` is declared under no slug on any board, so it collides
+        with nothing.
+
+        Confirmed working on a PB1600PS1 and a PBV4PS2, which is better
+        evidence than the byte pattern.
         """
         return await self._send_hex_command("FE0101FF")
 
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -427,6 +427,33 @@ class PitBoss:
             return {}
         return await self._send_command("turn-light-off")
 
+    async def turn_grill_on(self) -> dict:
+        """Lights the grill.
+
+        **This starts a fire in an appliance nobody may be standing next to.**
+        The board accepts it whatever state the grill is in, so anything built
+        on this should decide for itself whether starting is a good idea --
+        the library only sends it.
+
+        Sent as a raw MCU command because no board declares a slug for it. All
+        137 models declare `turn-off`; **not one declares a `turn-on`**, which
+        looks deliberate on the vendor's part rather than an oversight. The
+        byte layout leaves no doubt about which command it is:
+
+        ==================  ==========
+        ``turn-off``        FE0102FF
+        ``turn-light-on``   FE0201FF
+        ``turn-light-off``  FE0202FF
+        this                FE0101FF
+        ==================  ==========
+
+        The third byte is the subsystem -- 01 power, 02 light -- and the
+        fourth is on or off, so this is the missing member of a pair the board
+        already implements three quarters of. Confirmed working on a
+        PB1600PS1 and a PBV4PS2.
+        """
+        return await self._send_hex_command("FE0101FF")
+
     async def turn_grill_off(self) -> dict:
         """Turns the grill off."""
         return await self._send_command("turn-off")
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -850,3 +850,18 @@ async def test_no_board_declares_a_turn_on_command():
 
     assert with_off > 0
     assert with_on == 0
+
+
+def test_the_turn_on_command_collides_with_nothing():
+    """`FE0101FF` must not be some other board's command under a slug.
+
+    The safety property behind sending a raw command: whatever else the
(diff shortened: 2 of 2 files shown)


dknowles2 (commented):
The change is right, the behaviour is right, and it is the method David asked for on ha-pitboss#322 — I checked that rather than take it from the description, and his comment there is explicit: he tested the command on his PBV4PS2 and asked for a pytboss method before anything downstream. So this is not speculative surface.

I verified every catalogue claim by executing it:

```
turn-off       : 137 of 137
turn-on        :   0 of 137      (also no `power-on`, no `start`)
turn-light-on  : 132 of 137
turn-light-off : 132 of 137
```

And the thing I most wanted to know before a raw hex command goes into the public API: **`FE0101FF` is not declared under any slug on any board**, and the only fixed `FE01xxFF` command in the whole catalogue is `FE0102FF` for `turn-off`. Nothing collides.

The test that pins "no board declares `turn-on`" against the real catalogue is the right instinct — that claim is load-bearing for why this is raw hex, and a definitions refresh could quietly invalidate it.

## One correction, and it is the same trap as #541

> I generated those three from `grills.json` rather than reading them off a datasheet.

**Then the table would have shown two values for `turn-light-off`, and it shows one.**

```
FE0200FF -> LBL LFS PBA PBB PBC PBC2 PBD PBE PBG PBM PBM2 PBT PBV   (121 models)
FE0202FF -> PBL PBL3 PBP                                            ( 11 models)
```

The docstring lists `FE0202FF`, which is the **minority** variant — 11 of 132. It is also the one your PB1600PS1 and David's PBV4PS2 both use, since both are PBL-family, so this looks like it came from the boards in front of you rather than from the catalogue.

That matters because it makes the stated rule tidier than the data is. "The fourth byte is on or off" holds for power (`02` = off on all 137) and for light-on (`01`), but most boards use `00` for light-off, not `02`. So the fourth byte is not a uniform flag across subsystems — `01` means on, and off is spelled `02` or `00` depending on where you look.

**None of that weakens the conclusion.** `01` is still the only value that ever means "on", it is still the missing member of the pair, and it is confirmed on two grills — which is stronger evidence than the byte pattern anyway. But the docstring will be read as the derivation by whoever later wonders why a raw hex command lives in the API, and it should not present a cleaner pattern than exists. Either add the `FE0200FF` row, or say the fourth byte is `01` for on and varies for off.

`turn-off` itself is uniform — `FE0102FF` on all 17 boards — so that row is exactly right.

## On the deliberate absence of safety logic

Agreed, and the docstring says the necessary thing plainly rather than hedging: it starts a fire, the board accepts it in any state, deciding whether that is wise belongs to the caller. Putting the gating in the integration — the error flags, already-on, disconnected, plus an opt-in — is the right split. A transport library that silently refused would be worse, because a caller cannot tell refusal from failure.

754 tests, ruff, `ruff format --check` and mypy clean. Labelled `enhancement`.

Happy to merge as soon as the table is accurate; say if you would rather land it and fix the docstring separately.

trailro:
Corrected at `2efce0f`, and you caught exactly what happened.

**I generated the table from one grill, not from the catalogue.** `grills.get_grill("PB1600PS1")` — a PBL board — and then wrote "I generated those three from `grills.json`". Running it across all 137 gives what you got:

```
turn-off        FE0102FF   137
turn-light-on   FE0201FF   132
turn-light-off  FE0200FF   121
turn-light-off  FE0202FF    11
```

So I picked the minority spelling and presented it as the rule, and the grill it came from is the same family as both of the ones we tested on. That is the #541 trap again — one board's data read as the catalogue's — and the sentence claiming provenance is what made it look checked.

The docstring now carries all four rows with counts, and states the narrower claim that actually holds: **the fourth byte is `01` wherever a command means on; off is spelled `02` or `00` depending on the board, so it is not a uniform flag.** The tidy version was tidier than the data.

**Added the test for the property you went looking for**, since it is the real safety argument and was only in your review:

```python
def test_the_turn_on_command_collides_with_nothing():
    for grill in grills.get_grills():
        for slug, command in grill.control_board.commands.items():
            assert command() != "FE0101FF", f"{grill.name} declares it as {slug}"
```

1358 fixed commands across the catalogue, none of them this. That is a better guarantee than the byte pattern, and unlike the pattern it fails loudly if a definitions refresh ever introduces a collision.

The `turn-on` absence test stays as it was. Between the two, both claims the docstring rests on are now pinned against real data rather than prose.

**On landing it versus fixing separately:** I would rather it went in accurate. The docstring is the only place anyone will look to understand why a raw hex command is in the public API, and a wrong derivation there is worse than no derivation.

755 tests, ruff, `ruff format --check` and mypy clean.

dknowles2 (approved):
Every row in the corrected table matches the catalogue exactly — I re-derived all four rather than checking the two that changed:

```
turn-off        FE0102FF  137
turn-light-on   FE0201FF  132
turn-light-off  FE0200FF  121
turn-light-off  FE0202FF   11
```

And the narrower claim is the one the data supports: `01` is the only value that ever means on, off is spelled two ways, so the fourth byte is not a uniform flag. That is less satisfying than the original and it is what is true.

**Tracing it back to `get_grill("PB1600PS1")` is the part worth having.** The error was not the wrong hex — it was writing "I generated those three from `grills.json`" over something generated from one board. That sentence is what made it look checked, and it is why I went and executed it rather than reading past it. A provenance claim is a claim.

The extra sting is that the board you sampled is the same family as both grills the command was tested on, so nothing about the hardware evidence would have caught it either.

**Promoting the collision check into a test is the right move and it is the better guarantee.** The byte pattern is an argument; `FE0101FF` appearing under no slug anywhere is a fact, and now one that fails loudly if a definitions refresh introduces a collision. I had confirmed it in review, which protects today and nothing after.

> I would rather it went in accurate. The docstring is the only place anyone will look to understand why a raw hex command is in the public API.

Agreed, and that is the right reason. A wrong derivation there is worse than no derivation, because it stops the next person looking.

755 tests, ruff, `ruff format --check` and mypy clean; CI green. Merging.

This is now the method David asked for on ha-pitboss#322, with both of the claims it rests on pinned against real data. The Home Assistant side stays where it is until this is in a release, as you said.

PR #551, opened 2026-08-03
Annotate send_command as returning dict or None

### LINKED ISSUE #542: Transport.send_command is annotated -> dict but returns None for result-less handlers

trailro:
Fixes #542, taking option 1 — the union is the true type, per the firmware audit posted there.

## What the vendor images established

Across all seven mJS versions (0.2.3–0.6.0): **every setter handler the library wraps returns `null` on success in every version where it exists** — `PB.SendMCUCommand` included, which sits under every MCU command wrapper in `api.py` — and **every getter answers with an object in every version**. So the split is clean and universally safe:

- **`Transport.send_command` → `dict[Any, Any] | None`.** The `SendCommandFn` alias a few lines above already declared exactly this union; only the method's own annotation disagreed.
- **Setter wrappers** (`set_grill_temperature`, `set_probe_*`, `turn_*`, `set_temperature_unit`, `set_virtual_data`, `set_mcu_update_timer`, `set_wifi_update_frequency`, `Config.set`/`set_wifi_*`, `FS.Put`/`Rename`/`Remove`, `ping`) → `dict | None`. On real firmware these have been resolving `None` all along; the fake returning `{}` was what the tests were asserting.
- **Getters narrow at their own call sites** (`get_state`, `get_uptime`, `get_firmware_version`, `get_file_content`, `Config.get_info`, `get_config`) — their handlers answer with an object in every image, so `None` there only ever means a broken exchange, and callers keep the `dict` contract they had.

## The test you called for

`FakeTransport._send_prepared_command` now omits the `result` key when a handler returns nothing — the exact shape #534 fixed — and `_send_mcu_command` / `_set_virtual_data` / `_set_password` return `None` to match their firmware counterparts (`PB.SetVirtualData`'s success path has no return statement at all). Every MCU-command test now drives a result-less reply through the real request/response machinery and asserts `is None`; mocks returning `{}` on setter paths are gone.

## Compatibility

Runtime behavior is unchanged everywhere — this is annotations plus boundary narrowing. ha-pitboss consumes exactly four return values (`get_firmware_version`, `config.get_info`, `get_virtual_data`, `get_state`); all four are getters and keep returning a guaranteed dict. Setter results are not consumed there.

## Testing

`pytest` (751 passed), `ruff check`, `ruff format --check`, `mypy .` green at the CI-pinned versions.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -217,12 +217,12 @@ class PitBoss:
         psw = encode(self._password, key=timed_key(await self.get_uptime())).hex()
         return {**params, "psw": psw}
 
-    async def _send_hex_command(self, cmd: str) -> dict:
+    async def _send_hex_command(self, cmd: str) -> dict | None:
         return await self._conn.send_command(
             "PB.SendMCUCommand", await self._authenticate({"command": cmd})
         )
 
-    async def _send_command(self, slug: str, *args) -> dict:
+    async def _send_command(self, slug: str, *args) -> dict | None:
         cmd = self.spec.control_board.commands[slug]
         return await self._send_hex_command(cmd(*args))
 
@@ -254,7 +254,7 @@ class PitBoss:
         # they floor, so 190F is 87C to the grill rather than 88.
         return [floor((v - 32) / 1.8) for v in increments]
 
-    async def set_grill_temperature(self, temp: int) -> dict:
+    async def set_grill_temperature(self, temp: int) -> dict | None:
         """Sets the target grill temperature.
 
         Snapped to the nearest value the control board accepts, expressed in
@@ -273,7 +273,7 @@ class PitBoss:
         # ignore the extra argument, as do fixed-hex commands.
         return await self._send_command("set-temperature", temp, fahrenheit)
 
-    async def set_probe_temperature(self, temp: int) -> dict:
+    async def set_probe_temperature(self, temp: int) -> dict | None:
         """Sets the target temperature for probe 1.
 
         :param temp: Target probe temperature, in the grill's own unit.
@@ -282,7 +282,7 @@ class PitBoss:
             "set-probe-1-temperature", temp, self._is_fahrenheit()
         )
 
-    async def set_probe_2_temperature(self, temp: int) -> dict:
+    async def set_probe_2_temperature(self, temp: int) -> dict | None:
         """Sets the target temperature for probe 2.
 
         :param temp: Target probe temperature, in the grill's own unit.
@@ -403,7 +403,7 @@ class PitBoss:
         """
         await self._conn.send_command("PB.WiFiAwakeWDT", await self._authenticate({}))
 
-    async def set_temperature_unit(self, fahrenheit: bool) -> dict:
+    async def set_temperature_unit(self, fahrenheit: bool) -> dict | None:
         """Switches the unit the grill itself works in.
 
         This is the grill's own setting -- the one shown on its panel and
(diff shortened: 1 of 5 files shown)


dknowles2 (approved):
This closes #542, which I filed against ourselves, and it is a better answer than the one I proposed there. I suggested `dict | None` and left it at that; the useful part is the audit that decides *where* the union stops.

**I corroborated the setter/getter split against `fake_firmware/init.js` for every handler the library wraps**, by walking the handler bodies rather than reading them:

```
GetFirmwareVersion  object     SendMCUCommand          null
GetState            object     SetWiFiUpdateFrequency  null
GetVirtualData      object     SetMCU_UpdateFrequency  null
GetTime             object     WiFiAwakeWDT            null
                               SetVirtualData          null
                               SetDevicePassword       null
                               SetWifiCredentials      null
```

Clean split, exactly as you describe. **`PB.RenameDevice` is the one exception** — a setter that answers with an object — but the library does not wrap it, so it does not bear on this. Worth knowing the pattern is not universal in case someone later reasons "setters return null" about an RPC we have not wrapped yet.

The usual caveat, stated because it keeps mattering: that is the repo's model, not the firmware. You read all seven images and I read one file. Where they agree I am confident; the versions I cannot see rest on your audit.

## The thing I went looking for and did not find

I checked whether narrowing `_send_hex_command` would collide with #550's `turn_grill_on() -> dict`, which landed after this branched — merging `main` into the old head gave a conflict in `api.py` and a mypy break. **You had already rebased and narrowed it to `dict | None` before I got there.** Verified at `f394908`.

I also audited every wrapper's annotation against whether it is a getter or a setter, and they line up — the only flag was `_authenticate`, which builds its own dict and is correctly `-> dict`.

## Compatibility

Confirmed against ha-pitboss `main`: the consumed return values are `get_firmware_version`, `config.get_info`, `get_state` and **`get_probe_targets`** — not `get_virtual_data`, which #341 moved behind the new API. All four still return a guaranteed dict, so nothing downstream needs a change. Small correction to your list; it does not affect the conclusion.

**The test change is the part that makes this stick.** `FakeTransport` omitting `result` for handlers that return nothing means the setter paths now drive a genuinely result-less reply through the real request/response machinery. That is what was missing when I filed it — mocks returning `{}` were asserting a shape no grill sends, which is exactly how the wrong annotation survived #534.

755 tests, ruff, `ruff format --check` and mypy clean; CI green. Labelled `bug`. Merging.

PR #553, opened 2026-08-03
Serialize the BLE connection lifecycle and coalesce duplicate resets

### LINKED ISSUE #552: BLE: concurrent reset_device calls interleave and can leak a connected client
The last lifecycle item left over from #540's family, found while reviewing ha-pitboss's BLE path — filing it separately since #540 is closed and this one was not on its list.

`BleConnection.reset_device` runs a disconnect/connect pair with nothing serializing it — `_reconnecting` only suppresses the disconnect callback. `connect()` guards on `self._is_connected`, but that flag is False for the whole time `establish_connection` is working, which is seconds under bleak-retry.

That window is exactly where concurrent resets come from in practice. ha-pitboss's discovery callback schedules `reset_device(service_info.device)` on **every advertisement** while `conn.is_connected()` is False (`_connect_ble` in its `__init__.py`), and PitBoss grills advertise every second or two. During one connect attempt, several more resets are queued; each starts with `await self.disconnect()` — tearing down whatever the previous task just built — and then races its own `establish_connection`.

Interleavings that follow from the code as written:

- Task B's `disconnect()` runs while task A is inside `establish_connection`; A's client lands afterwards and is assigned over B's — whichever loses stays **connected with no reference held**, leaked until the grill drops it.
- `start_notify` from one task races the other task's disconnect, surfacing as `BleakError` from inside a reset that should have succeeded.
- `_ble_device` and `_ble_client` can end up from different resets.

Symptoms at the consumer: connect/disconnect churn during setup, doubled GATT connections against a grill that only serves one, and sporadic `Failed to connect` noise on a grill that is right there advertising.

Two-part fix, PR to follow:

1. A lifecycle lock serializing `connect` / `disconnect` / `reset_device`. Distinct from `self._lock`, which serializes GATT I/O — holding that one across a multi-second connect would stall the receive path.
2. Serialization alone still runs every queued reset, each tearing down the connection the previous one just built against the same grill. So a reset that finds the transport already connected to the **same address** only adopts the fresher `BLEDevice` and returns.

The consumer-side half (don't schedule a reset while one is in flight) stops mattering once the transport is safe under concurrent calls, which is the right place to fix it — any discovery-driven caller hits this, not just ha-pitboss.

trailro:
Fixes #552.

A `_lifecycle_lock` now serializes `connect`, `disconnect` and `reset_device`; the public methods take the lock and delegate to `_connect_locked` / `_disconnect_locked`, and `reset_device` holds it once across its whole disconnect/connect pair. It is deliberately distinct from `self._lock` (GATT I/O): holding that one for the length of an `establish_connection` would stall the receive path.

Serialization alone would still run every queued reset in turn, each starting with a `disconnect()` that tears down the connection the previous one just built against the same grill. So a reset that finds the transport already connected to the same **address** now only adopts the fresher `BLEDevice` — advertisements carry updated metadata worth keeping — and returns without touching the connection. A reset for a *different* address still does the full teardown; the coalescing key is the address, nothing else.

Behaviour otherwise unchanged: `_reconnecting` still only suppresses the disconnect callback and is still cleared in a `finally`, the failed-connect path still propagates (the existing regression test for it passes untouched), and `_on_disconnected` stays lock-free — it is a sync bleak callback and takes no part in the lifecycle.

Three new tests: concurrent resets never overlap inside `establish_connection` (tracked via an in-flight counter in the mock), queued duplicate resets for the same address land one `establish_connection` and zero disconnects while still adopting the newer `BLEDevice`, and a reset to a different address still reconnects fully.

Rebased on current `main` (post #547/#549/#551 — this composes with #547's hygiene fixes rather than overlapping them; none of its lines are touched). 758 tests, `ruff check`, `ruff format --check` and `mypy .` clean. Not exercised on hardware: my grill runs the local WiFi transport these days, so the race analysis and the interleaving are established by the code reading in #552 and pinned down deterministically by the in-flight-counter test, not by a live BLE session.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -64,12 +64,26 @@ class BleConnection(Transport):
         self._disconnect_callback = disconnect_callback
         self._is_connected = False
         self._reconnecting = False
+        # Serializes connect/disconnect/reset_device. Establishing a
+        # connection takes seconds, and a consumer driving reconnects from
+        # discovery -- Home Assistant schedules a reset per advertisement
+        # while `is_connected()` is False -- piles several up in exactly that
+        # window. Unserialized, their disconnect/connect steps interleave on
+        # the same object and the losing `establish_connection` leaks a
+        # connected client. Distinct from `self._lock`, which serializes GATT
+        # I/O; holding that one for the length of a connect would stall the
+        # receive path.
+        self._lifecycle_lock = asyncio.Lock()
 
     async def connect(self) -> None:
         """Starts the connection to the device.
 
         Does nothing if already connected or if no BLE device was set.
         """
+        async with self._lifecycle_lock:
+            await self._connect_locked()
+
+    async def _connect_locked(self) -> None:
         if self._is_connected:
             _LOGGER.warning("Already connected. Ignoring call to connect().")
             return
@@ -110,6 +124,10 @@ class BleConnection(Transport):
 
     async def disconnect(self) -> None:
         """Stops the connection to the device."""
+        async with self._lifecycle_lock:
+            await self._disconnect_locked()
+
+    async def _disconnect_locked(self) -> None:
         _LOGGER.debug("Disconnecting from device.")
         if self._ble_client:
             try:
@@ -126,17 +144,32 @@ class BleConnection(Transport):
     async def reset_device(self, ble_device: BLEDevice):
         """Resets the BLE device used for transport.
 
+        A reset that arrives while already connected to the same device --
+        by address -- only adopts the fresher `BLEDevice` and keeps the
+        connection. Discovery-driven consumers queue a reset per
+        advertisement seen while disconnected, and by the time the later
+        ones run, the first has already connected; tearing that connection
+        down to rebuild it against the same grill was pure churn.
+
         :param ble_device: BLE device to use for transport.
         """
-        self._reconnecting = True
-        _LOGGER.debug("Resetting device to: %s", ble_device)
-        try:
-            await self.disconnect()
-            self._is_connected = False
-            self._ble_device = ble_device
-            await self.connect()
-        finally:
-            self._reconnecting = False
+        async with self._lifecycle_lock:
+            if (
+                self._is_connected
+                and self._ble_device is not None
+                and self._ble_device.address == ble_device.address
+            ):
+                _LOGGER.debug("Already connected to %s; keeping it", ble_device)
+                self._ble_device = ble_device
+                return
+            self._reconnecting = True
+            _LOGGER.debug("Resetting device to: %s", ble_device)
+            try:
+                await self._disconnect_locked()
+                self._ble_device = ble_device
+                await self._connect_locked()
+            finally:
+                self._reconnecting = False
 
     def is_connected(self) -> bool:
         """Whether the device is currently connected."""
--- a/tests/test_ble.py
+++ b/tests/test_ble.py
@@ -118,6 +118,91 @@ async def test_reset_device_connect_failure_resets_reconnecting_flag(
 
 
 @mock.patch("bleak_retry_connector.establish_connection")
+async def test_concurrent_resets_run_one_at_a_time(mock_establish_connection):
+    """Two resets racing must not interleave their disconnect/connect steps.
+
+    Establishing a connection takes seconds, and a consumer driving resets
+    from discovery schedules one per advertisement seen while disconnected --
+    so several are queued in exactly that window. Unserialized, the losing
+    `establish_connection` leaked a connected client.
+    """
+    device_a = mock.create_autospec(bleak.BLEDevice)
+    device_b = mock.create_autospec(bleak.BLEDevice)
+    device_a.name = device_b.name = "GRILL"
+
+    in_flight = 0
+    max_in_flight = 0
+
+    async def slow_establish(**kwargs):
+        nonlocal in_flight, max_in_flight
+        in_flight += 1
+        max_in_flight = max(max_in_flight, in_flight)
+        # Yield so the racing reset gets its chance to interleave.
+        await asyncio.sleep(0)
+        await asyncio.sleep(0)
+        in_flight -= 1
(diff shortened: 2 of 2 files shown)


dknowles2 (commented):
The analysis is right and I confirmed the caller behaves exactly as you describe. `ha-pitboss`'s `_detection_callback` returns early `if conn.is_connected()`, then does `entry.async_create_task(...)` — so it never *schedules* a reset while connected, but several can be queued before the first completes, and by the time the later ones run the connection exists. That is precisely the case the coalescing targets, and it means the coalescing cannot suppress a reset anyone deliberately asked for.

Separating `_lifecycle_lock` from `self._lock` is the right call for the reason you give — holding the GATT lock across an `establish_connection` would stall the receive path for seconds.

Dropping the redundant `self._is_connected = False` from `reset_device` is correct too: `_disconnect_locked` sets it as its last statement, so the explicit line was already dead. Worth that it went rather than being carried along.

## What the tests actually pin, which is less than the title claims

I mutated each lock acquisition separately:

```
remove the lock from reset_device  -> test_concurrent_resets_run_one_at_a_time   FAILS
remove the coalescing             -> test_a_queued_duplicate_reset_keeps_...     FAILS
remove the lock from connect()    -> 25 passed
remove the lock from disconnect() -> 25 passed
```

So two of the four acquisitions are unpinned. Every new test drives `reset_device`, which takes the lock itself, so the ones on `connect()` and `disconnect()` are load-bearing only for callers that reach them directly — and nothing would notice if a later reader decided they were redundant because "`reset_device` already locks".

**And that path is reachable, narrowly.** In `_connect_ble` the connection is established entirely through `reset_device`; `api.start()` calls `conn.connect()` afterwards. If the link drops in the window between `_connect_ble` returning and `start()` running, `_is_connected` goes False, the detection callback queues a `reset_device`, and `connect()` runs against it concurrently. That is exactly what the lock on `connect()` is for, and it is the one interleaving no test covers.

A test in the shape of the one you already wrote — the in-flight counter, but racing `connect()` against `reset_device` rather than two resets — would close it. Your counter mock is already the hard part.

## One consequence worth recording

Coalescing removes an accidental recovery path. Before this, a queued duplicate tore the connection down and rebuilt it, which would incidentally revive a link that was wedged while `_is_connected` still read True. After, it returns early. Nobody was relying on that deliberately and the churn it caused is the bug being fixed, so I think the trade is right — but if wedged-connection reports ever appear, this is the change to look at first.

## Verified

758 tests, ruff, `ruff format --check`, mypy clean; CI green. Rebased on current `main` and it does compose with #547 rather than overlapping — none of those lines move.

Saying plainly that this is not exercised on hardware, and why (your grill runs the local transport now), is the right way to present it. The in-flight counter makes the interleaving deterministic, which is better evidence for a race than a live session would be anyway.

I have labelled #552 `bug` and this PR the same. Happy to merge as soon as `connect()` has a test, or to land it as-is if you would rather follow up — your call, it is a coverage gap rather than a defect.

trailro:
Test added: `test_connect_racing_a_reset_runs_one_at_a_time` — the in-flight counter shape, racing `connect()` against `reset_device()` on a disconnected transport, with the reachable interleaving you traced (`_connect_ble` connects via reset, the link drops before `api.start()`'s `connect()` runs, the drop queues another reset) recorded in its docstring so the next reader knows why the lock on `connect()` is load-bearing.

Verified the same way you found the gap: with the lock removed from `connect()`, the new test fails; with it, 759 tests pass. I left `disconnect()`'s acquisition unpinned deliberately — every interleaving I could construct for it on a real call path goes through `reset_device` or ends with the transport disconnected either way, so a test would pin the implementation rather than an observable outcome. If you can see a reachable one, happy to add it.

Also agreed on recording the coalescing trade-off: a queued duplicate no longer incidentally revives a link wedged with `_is_connected` stuck True. Your framing — the change to look at first if wedged-connection reports appear — is exactly what #552 should carry, so I have noted it there.

ruff, `ruff format --check`, mypy clean.

dknowles2 (approved):
`connect()` is pinned now — I confirmed it the same way, removing the lock from `connect()` alone and watching `test_connect_racing_a_reset_runs_one_at_a_time` fail. 759 tests, ruff, format and mypy clean; CI green. Merging.

## You asked for a reachable interleaving for `disconnect()`. There is one, and it leaks

Your argument was fair — every path you could construct went through `reset_device` or ended disconnected either way. The one that does not is **entry unload racing a queued reset on a link that has just dropped**:

1. Discovery connects through `reset_device`.
2. The link drops; `_on_disconnected` clears `_is_connected`.
3. The drop queues another `reset_device` — and because the flag is now False, it does the full teardown/reconnect rather than coalescing.
4. The entry unloads: `api.stop()` calls `disconnect()` while that reset is mid-flight.

Ran both ways, same scenario:

```
with the lock:     established=2  is_connected=False  client_set=False  never_disconnected=0
without the lock:  established=2  is_connected=True   client_set=True   never_disconnected=1
```

Without it, `disconnect()` slots between reset's teardown and its reconnect, so the client reset establishes afterwards survives the unload — **a live bleak connection nobody disconnected, on a transport that reports connected after an explicit `disconnect()`**. That is #552's leak reached through a different pair.

Both of those are observable outcomes rather than implementation details: `is_connected()` returning True after `disconnect()` returns, and a client whose `disconnect()` was never called. So I think it is pinnable without over-fitting — the scenario above is the whole test, and your existing counter mock is not even needed for it.

**Not holding the merge for it.** The lock is already there and correct; this is coverage, and I said I would merge once `connect()` was covered rather than moving the line afterwards. Worth a follow-up when convenient, or leave it and it is still right today.

Agreed on carrying the coalescing trade-off into #552 rather than the PR description — that is where someone chasing a wedged-connection report will look.

One thing I got wrong earlier and should say plainly: when I first mutated the lock I removed it from `connect()` and reported that as "unpinned", which was accurate but incomplete — I had not yet checked whether the path was reachable, and only worked that out afterwards from `_connect_ble`. The order should have been the other way round: establish reachability, then report the gap.

PR #554, opened 2026-08-03
Log the already-connected connect() no-op at debug level

trailro:
One line (plus its rationale as a comment): the "Already connected. Ignoring call to connect()." log drops from warning to debug.

It fires on a healthy, documented path. ha-pitboss establishes the Bluetooth connection through `reset_device` during discovery and then calls `PitBoss.start()`, whose `connect()` lands on this guard — so every successful Bluetooth setup emits a warning about behaviour that is working as designed. A warning that always fires teaches people to ignore warnings; the guard itself is correct and stays.

Stacked on #553 (`serialize-ble-lifecycle`), where the guard now lives in `_connect_locked` — the diff includes that PR until it lands and reduces to this one line after. Happy to rebase either way once #553 settles.

759 tests, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -85,7 +85,11 @@ class BleConnection(Transport):
 
     async def _connect_locked(self) -> None:
         if self._is_connected:
-            _LOGGER.warning("Already connected. Ignoring call to connect().")
+            # Debug, not a warning: this is a documented no-op, and consumers
+            # hit it on a normal path -- Home Assistant connects through
+            # `reset_device` during discovery and then calls `PitBoss.start`,
+            # whose `connect()` lands here at every Bluetooth setup.
+            _LOGGER.debug("Already connected. Ignoring call to connect().")
             return
         if self._ble_device is None:
             return
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -64,12 +64,26 @@ class BleConnection(Transport):
         self._disconnect_callback = disconnect_callback
         self._is_connected = False
         self._reconnecting = False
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Correct, and I can confirm the path from my own reading rather than yours — I traced exactly this while reviewing #553. `_connect_ble` establishes the link entirely through `reset_device` and waits on `ready.wait_for(conn.is_connected)`; `PitBoss.start()` then calls `conn.connect()`, which lands on the already-connected guard. So every successful Bluetooth setup has been emitting a warning about working as designed.

> A warning that always fires teaches people to ignore warnings.

That is the right reason, and it is worth more than the log level itself. The guard stays because it is correct; only the claim that something is wrong goes away.

Now that #553 has landed, this reduces to what you said it would — I checked against current `main` rather than the PR's own diff, which still shows #553's content because the merge base predates it:

```
-            _LOGGER.warning("Already connected. Ignoring call to connect().")
+            # Debug, not a warning: ...
+            _LOGGER.debug("Already connected. Ignoring call to connect().")
```

One line and a comment, plus a comment fix in `test_connect_when_already_connected`. No conflicts.

759 tests, ruff, `ruff format --check` and mypy clean; CI green. Merging.

One observation, not a request: nothing pins the level, so the change is invisible to the suite — `test_connect_when_already_connected` asserts the no-op and its comment now says "debug", but a future edit back to `warning` would pass. A `caplog` assertion would close that, though I can see the argument that pinning log levels over-fits. Your call; I would probably leave it.

PR #556, opened 2026-08-03
Base64-encode the data sent to FS.Put

### LINKED ISSUE #555: FS.Put sends data unencoded, but the RPC expects base64
`FileSystem.set_file_content()` passes `data` to `FS.Put` as-is:

https://github.com/dknowles2/pytboss/blob/main/pytboss/fs.py#L57-L60

```python
return await self._conn.send_command(
    "FS.Put",
    {"filename": filename, "data": data, "append": append},
)
```

but the parameter is base64. The device says so itself — `RPC.Describe` against a PB1600PS1 on 0.5.7:

```
FS.Put: {"name": "FS.Put", "args_fmt": "{filename: %Q, offset: %ld, data: %V, append: %B}"}
FS.Get: {"name": "FS.Get", "args_fmt": "{filename: %Q, offset: %ld, len: %ld}"}
```

`%V` is Mongoose's format specifier for base64-encoded binary, as against the `%Q` it uses for the plain string in the very same signature. The asymmetry is visible inside the module too: `get_file_content()` already `b64decode`s what `FS.Get` returns, so the two halves of the same file transfer disagree about the encoding.

The consequence is a silent one, which is what makes it worth fixing rather than documenting. The firmware base64-decodes whatever arrives, so writing `"hello"` does not fail — it stores the bytes `\x85\xe9e`. Anything whose text happens to be valid base64 lands as something else entirely, and anything that is not decodes to a shorter mangled blob. A file written this way and read back with `get_file_content()` does not match what was sent.

Two smaller things noticed alongside it, neither of which I would fold into a fix:

* `FS.Put` accepts an `offset` the wrapper does not expose. `append` covers the sequential case, so this is a completeness note rather than a gap.
* A whole file goes out in a single RPC. Base64 inflates by a third and the RPC frame is bounded, so large writes want chunking via `append` — a caller's concern, but worth knowing before someone tries to write a 15 kB file.

PR up shortly.

trailro:
Fixes #555.

## The evidence

`RPC.Describe` against a PB1600PS1 on 0.5.7:

```
FS.Put: {"name": "FS.Put", "args_fmt": "{filename: %Q, offset: %ld, data: %V, append: %B}"}
```

`%V` is base64-encoded binary; `%Q`, in the same signature, is the plain string. So `data` is base64 and `filename` is not, which is also what the module's own read path assumed — `get_file_content()` has always `b64decode`d what `FS.Get` returns.

The failure is silent rather than loud: the firmware decodes whatever arrives, so writing `"hello"` stored `\x85\xe9e` instead of failing. A file written and then read back did not match.

## The change

Encode in `set_file_content()`, and accept `bytes` as well as `str` — the encoding now happens inside the wrapper, so a caller with binary content previously had no way to express it. Text is UTF-8 encoded first, matching what the read path decodes.

## Testing

Three tests: the request shape, a **round trip** (`set_file_content` → the bytes it sends → `get_file_content`) over multi-byte text that a raw write would corrupt, and a bytes payload. The round trip is the one with teeth — it fails on the old code, since that is precisely the property that was broken.

761 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

## Not verified on hardware

I have a grill that answers `FS.Put`, but I have not written to its filesystem to confirm — that is its running configuration and mJS app, and a botched write is not worth the confirmation. The `%V` declaration comes from the device rather than from the docs, which seems like the stronger evidence anyway. Happy to do a scratch-file round trip against it if you would rather have that too.

No downstream impact: neither ha-pitboss nor any integration I know of calls `set_file_content`.

### DIFF
--- a/pytboss/fs.py
+++ b/pytboss/fs.py
@@ -1,6 +1,6 @@
 """Client library for Mongoose OS filesystem RPCs."""
 
-from base64 import b64decode
+from base64 import b64decode, b64encode
 
 from .transport import Transport
 
@@ -45,18 +45,33 @@ class FileSystem:
                 return content.decode("utf-8")
 
     async def set_file_content(
-        self, filename: str, data: str, append: bool
+        self, filename: str, data: str | bytes, append: bool
     ) -> dict | None:
         """Writes content to a file on the device.
 
+        `data` is base64-encoded before it is sent, which is what the RPC
+        expects: the device describes the parameter as `%V` -- Mongoose's
+        format specifier for base64-encoded binary -- rather than the `%Q` it
+        uses for the plain string alongside it::
+
+            FS.Put  {filename: %Q, offset: %ld, data: %V, append: %B}
+
+        which is the same encoding `get_file_content()` decodes on the way
+        back.
+
         :param filename: Path of the file to write.
-        :param data: Content to write.
+        :param data: Content to write. Text is encoded as UTF-8.
         :param append: If True, appends to the existing file instead of
             overwriting it.
         """
+        raw = data.encode("utf-8") if isinstance(data, str) else data
         return await self._conn.send_command(
             "FS.Put",
-            {"filename": filename, "data": data, "append": append},
+            {
+                "filename": filename,
+                "data": b64encode(raw).decode("ascii"),
+                "append": append,
+            },
         )
 
     async def rename_file(self, src: str, dst: str) -> dict | None:
--- a/tests/test_fs.py
+++ b/tests/test_fs.py
@@ -1,4 +1,4 @@
-from base64 import b64encode
+from base64 import b64decode, b64encode
 from unittest import mock
 from unittest.mock import AsyncMock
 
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
**I reproduced the corruption exactly, and the byte sequence in your description is right.** Mongoose decodes greedily and drops the trailing partial group, so writing `"hello"` decodes `"hell"`:

```
b64decode("hell") = b'\x85\xe9e'      matches your \x85\xe9e: True
```

Three bytes stored where five were meant, and nothing raises. That is the worst shape a bug can have — the write succeeds, the file exists, and only a read-back tells you.

**The decisive evidence for me is not `RPC.Describe`, it is that the library contradicted itself.** `get_file_content()` has always `b64decode`d what `FS.Get` returns. So one of our two paths was wrong regardless of what the device says, and the read path is the one that matches Mongoose's documented `%V`. `RPC.Describe` settles *which*, but the asymmetry alone establishes *that*.

Worth being explicit that I cannot corroborate the `%V` reading from this repo: **`fake_firmware/init.js` models no `FS.*` handlers at all** — they are Mongoose core RPCs the app never registers, the same gap `Sys.Reboot` and `OTA.*` have. So the `args_fmt` line comes from your grill and I am taking it on that basis.

## Verified

* 761 tests, ruff, `ruff format --check`, mypy clean; CI green.
* Mutated the encode away — three tests fail, including `test_set_file_content_round_trips_through_get`. That one is the one with teeth, as you say: it asserts the property that was actually broken rather than the shape of a request.
* Multi-byte text survives the round trip on the new path and is mangled on the old.

Accepting `bytes` as well as `str` is right and follows from the fix rather than being bundled with it — once the encoding moves inside the wrapper, a caller with binary content has no other way to express it. UTF-8 for text matches what the read path decodes, so the pair stays symmetrical.

**Not writing to your grill's filesystem to confirm is the correct call**, and I would have said so if you had asked first. Its running configuration and mJS app live there, and a botched write costs a reflash to learn something the device already told you in `args_fmt`. If you do want hardware confirmation later, a scratch path like `/tmp/pytboss-probe` would get it without going near anything that matters — but I do not think this needs it.

Labelled `bug`, and #555 the same. Merging.

PR #557, opened 2026-08-03
Pin the lifecycle lock on disconnect() with an unload-vs-reset race

trailro:
The follow-up from your review of #553: `disconnect()`'s lock acquisition was the last one unpinned, and you supplied the reachable interleaving I had failed to construct — discovery connects through `reset_device`, the link drops (clearing the flag and queueing a reset that now tears down and reconnects rather than coalescing), and the entry unload's `disconnect()` races that reset mid-flight.

Test-only; the scenario is the test, exactly as you laid it out, and as predicted the in-flight counter was not needed. The assertions are the observable outcomes rather than lock mechanics: after the gather, the transport reports disconnected, holds no client, and the client the reset established was handed to `disconnect()` instead of surviving the unload — your `never_disconnected=1` case. `establish_connection` count pins that the reset still did its full reconnect, so the test cannot silently pass by the reset never re-establishing.

Verified by mutation, same method as the `connect()` one: with the lock removed from `disconnect()` alone, this test fails on `is_connected()` reading True after `disconnect()` returned; with it, 760 tests pass. `ruff check`, `ruff format --check` and `mypy .` clean.

With this, all three public lifecycle entry points are individually pinned — `connect()` and `reset_device` from #553, `disconnect()` here — so none of the acquisitions can quietly become "redundant" in a refactor.

### DIFF
--- a/tests/test_ble.py
+++ b/tests/test_ble.py
@@ -191,6 +191,55 @@ async def test_connect_racing_a_reset_runs_one_at_a_time(mock_establish_connecti
 
 
 @mock.patch("bleak_retry_connector.establish_connection")
+async def test_disconnect_racing_a_reset_leaves_nothing_behind(
+    mock_establish_connection,
+):
+    """`disconnect()` takes the lifecycle lock in its own right too.
+
+    The reachable interleaving, from review of the lock: discovery connects
+    through `reset_device`; the link drops, so the flag is cleared and the
+    drop queues another reset -- which does the full teardown/reconnect
+    rather than coalescing; the entry then unloads, calling `disconnect()`
+    while that reset is mid-flight. Unserialized, `disconnect()` slots
+    between the reset's teardown and its reconnect: the client the reset
+    establishes afterwards survives the unload -- a live connection nobody
+    disconnected, on a transport reporting connected after an explicit
+    `disconnect()`.
+    """
+    first_seen = mock.create_autospec(bleak.BLEDevice)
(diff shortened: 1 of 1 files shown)


dknowles2 (approved):
Confirmed the same way you did — with the lock removed from `disconnect()` alone, this is the only test that fails:

```
FAILED tests/test_ble.py::test_disconnect_racing_a_reset_leaves_nothing_behind
1 failed, 26 passed
```

So it isolates that acquisition and nothing else, which is what makes it worth having.

**Asserting the observable outcomes rather than the lock is the right call**, and it is what I could not have told you from my probe alone — I had `is_connected`, `client_set` and `never_disconnected` as diagnostics; you turned them into the three things a caller would actually notice. Adding `await_count == 2` is the assertion I would have missed: without it the test passes trivially if the reset stops re-establishing, which is exactly how a race test rots into a tautology.

All three public entry points are individually pinned now. That was the point — not that the locks were wrong, but that two of the four acquisitions could have been deleted in a refactor as "redundant, `reset_device` already locks" with nothing objecting.

760 tests, ruff, `ruff format --check` and mypy clean; CI green. Labelled `bug`. Merging.

One note for the record, since this closes a thread that started with me getting the order wrong: on #553 I reported `connect()`'s lock as unpinned and only afterwards worked out from `_connect_ble` that the path was reachable. Reachability should come first — an unpinned guard on an unreachable path is not worth a round trip, and I did not know which I had when I raised it. It happened to be reachable both times.

PR #558, opened 2026-08-03
Gate the local endpoint on firmware family, not just http.enable

trailro:
Targets `http-transport` rather than `main`, same as #546 — merge it into #543 with one click, take the diff, or ignore it.

Follows the [observation on #543](https://github.com/dknowles2/pytboss/pull/543#issuecomment-5165950688). Docstring and tests only; no behaviour change.

## What the second firmware line changes

Dansons ships a line that is not Mongoose: C on ESP-IDF v4.4.x, versioned `16.x`, on PBC2/PBD/PBE/PBL2/PBT. Reading those images: `esp_http_server` is not linked (the `ESP_ERR_HTTPD_*` strings are `esp_err_to_name` entries, and the component's own source paths are absent where the websocket client's are present), the websocket is client-only, and there is no mDNS, no `bind`/`listen` in application code and no `/rpc` path. Those grills are reachable over Bluetooth and the relay, and by nothing else.

So availability is decided twice: by firmware family first, then by `http.enable` within the Mongoose family. Your per-unit framing was right for the family it describes — @dknowles2's PBV4PS2 and my PB1600PS1 are the same board and firmware with opposite values — it just is not the whole rule.

## Why that lands in the snippet

```python
http = await boss.config.get_config("http")
if http["enable"]:
```

assumes Mongoose's schema. The ESP-IDF firmware answers `Config.Get`, but its schema holds only `wifi.sta`, `app` and `version`, so there is no `http` section to report on. A device that will not answer for a key it does not have can either return an error or return nothing, so the snippet now handles both — `except RPCError` and `.get()` — and says why in a sentence each, since neither is obvious as defensive habit.

## Tests

The snippet is now **executed** rather than only rendered: `_documented_discovery()` in `tests/test_http.py` is the docstring's code verbatim, exercised against a grill that enables the endpoint, one that disables it, one that answers nothing, and one that errors on the key. Same reasoning as last time — documentation that gates a feature is code, and mine was wrong precisely because I never ran it.

## One note on the branch

`http-transport` predates #551, which narrowed `Config.get_config` to return `{}` rather than `None` for a result-less reply. Wrote the snippet against post-merge behaviour, so the empty-reply case is `{}`; if you merge `main` in first, nothing here needs revisiting.

## What is not verified

I do not have an ESP-IDF grill. This is static analysis of stripped release binaries — absence of a linked component is a stronger claim than absence of a string, but it is still not a device saying no. Anyone with a PBD/PBT/PBE/PBC2/PBL2 settles it in one request: if `http://<ip>/rpc` answers, I am wrong.

760 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -5,17 +5,40 @@ Dansons websocket relays, so the command mappings and status decoders in this
 library work against it unchanged. Talking to the grill directly removes the
 vendor's cloud from the path.
 
-**Not every grill answers.** The endpoint is gated by the firmware's
-`http.enable` setting, and that is per-unit configuration rather than
-something a model or firmware version predicts -- two grills on the same
-board and the same firmware have been observed with opposite values. Nothing
-here turns it on: `connect()` fails against a grill where it is off, and the
-caller is expected to have established that the endpoint exists first, over a
-transport that already works. `Config.Get` over Bluetooth reports it::
+**Not every grill answers**, and whether one does is decided twice over.
 
-    http = await boss.config.get_config("http")
-    if http["enable"]:
+First by firmware family. Dansons ships a second line that is not Mongoose at
+all -- C on ESP-IDF, versioned `16.x` rather than `0.x`, on the PBC2, PBD,
+PBE, PBL2 and PBT boards. Those images link no HTTP server and run their
+websocket as a client only, so they serve nothing on the local network: they
+are reachable over Bluetooth and the vendor's relay, and by nothing else. No
+setting changes that.
+
+Then, within the Mongoose family, by the firmware's `http.enable` setting.
+That part is per-unit configuration rather than something a model or firmware
+version predicts -- two grills on the same board and the same firmware have
+been observed with opposite values.
+
+Nothing here turns it on: `connect()` fails against a grill that does not
+serve the endpoint, whichever of the two reasons applies, and the caller is
+expected to have established that it exists first, over a transport that
+already works. `Config.Get` over Bluetooth reports it::
+
+    try:
+        http = await boss.config.get_config("http")
+    except RPCError:
+        http = {}  # no such section: not a Mongoose grill
+    if http.get("enable"):
         ...  # an HttpConnection will work against this grill
+
+Neither the `try` nor the `.get()` is defensive habit; each covers one of the
+two ways the question can fail to have an answer. The ESP-IDF firmware answers
+`Config.Get` too, but its schema holds only `wifi.sta`, `app` and `version`,
+so there is no `http` section to report on -- and a device that will not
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Taking this, with one thing to do on my side first and one caveat I want stated more sharply than the description does.

**Your branch note is right, and here is the consequence made concrete.** `http-transport` predates #551, so on this branch `Config.get_config` still returns `send_command(...)` unguarded, where `main` now has `or {}`. Running the documented snippet against all four reply shapes on *this branch*:

```
reply={'enable': True}   -> True
reply={'enable': False}  -> False
reply={}                 -> False
reply=None               -> AttributeError: 'NoneType' object has no attribute 'get'
```

The parametrised cases stop at `{}` deliberately, which is why the suite is green — and correctly so, since post-merge a result-less reply reaches the caller as `{}`. But until `main` is in, the snippet the docstring publishes is wrong for exactly the case it claims to cover. **So I will merge `main` into `http-transport` before this rather than after**, and then the `.get()` is doing what the sentence beside it says.

That is a good catch to have flagged rather than left for someone to trip over.

**Executing the snippet instead of only rendering it is the right lesson from #546**, and it is the second time that has paid. My original was wrong in two ways precisely because I wrote it to be copied and never ran it once.

## What I can check, and what I cannot

I sized the claim: **17 reachable models sit on PBC2, PBD, PBE and PBT** — PBT alone is 11 — and `PBL2` adds `PB1150DX` and `PB850DX` through the dual-board path that #354 opened up. So this is not a corner; it is about an eighth of the catalogue being unable to serve a local endpoint under any configuration.

And I cannot verify a word of the firmware analysis. There is **no trace of the ESP-IDF line anywhere in this repo** — `fake_firmware/init.js` models the Mongoose app only, and nothing in `grills.json` distinguishes a firmware family. So "no linked HTTP server, websocket client-only, no `/rpc`" rests entirely on your reading of stripped release binaries, and I am taking it on that basis rather than corroborating it.

Your own framing of that evidence is the right one and I want to endorse it explicitly: absence of a linked component is a stronger claim than absence of a string, and it is still not a device saying no. **One request from anyone with a PBD, PBT, PBE, PBC2 or a PBL2-board grill settles it** — and the docstring will be asserting something to every reader until someone makes it.

The two-stage framing is better than what I wrote. "Per-unit, no model predicts it" was true of the family I had looked at and I stated it as the whole rule; family-then-setting is the accurate shape.

760 tests, ruff, `ruff format --check` and mypy clean. No CI here — pytboss only runs it for PRs targeting `main`, so the local run is the check, which it can be for this repo.

Labelled `documentation`. Merging, then bringing `main` into the branch.

PR #559, opened 2026-08-03
Send PB.SetWiFiUpdateFrequency under the name the firmware registers

trailro:
`set_wifi_update_frequency()` asks for `PB.SetWifiUpdateFrequency`. Every firmware registers `PB.SetWiFiUpdateFrequency` — capital F. RPC names are matched exactly, so the call has never reached a grill.

## Confirmed on hardware

`RPC.Describe` against a PB1600PS1 on 0.5.7, both spellings:

```
PB.SetWifiUpdateFrequency -> {"code":404,"message":"name not found"}
PB.SetWiFiUpdateFrequency -> {"name":"PB.SetWiFiUpdateFrequency","args_fmt":""}
```

And in the vendor images, the capital F is consistent across the whole mJS line — 0.2.3, 0.4.0, 0.5.0, 0.5.5, 0.5.6, 0.5.7, 0.6.0 all contain `RPC.addHandler("PB.SetWiFiUpdateFrequency", ...)` and nothing else. There is no firmware where the library's spelling works.

The method's own docstring, two hundred lines up in `request_fast_updates()`, already refers to `PB.SetWiFiUpdateFrequency` correctly. The code and its documentation have simply disagreed.

## How it survived

The test asserted the name the code sent:

```python
send_command.assert_awaited_once_with(
    "PB.SetWifiUpdateFrequency", {"slow": 10, "fast": 1}
)
```

which passes for any spelling, including one no grill answers. Same shape of blind spot as #551's `{}`-returning mocks and #546's never-executed docstring.

So the test now drives `FakeTransport`, which registers the handler under the firmware's name and falls through to its unknown-method path otherwise — the fake refuses a wrong spelling the way a grill does. Reverting the one-character fix turns three tests red; with a mocked `send_command` it would still be green.

The second test pins the wire name directly, since that is the actual claim being made.

## Scope

One string in `api.py`. No signature or behaviour change beyond the call now arriving. Nothing downstream calls this method — ha-pitboss does not, and `request_fast_updates()` (#530), which shares the underlying mechanism, uses `PB.WiFiAwakeWDT` and is unaffected.

Worth noting for completeness: the ESP-IDF firmware line does not have this RPC at all, under either spelling — it registers only `PB.SetMCU_UpdateFrequency`. That is a separate matter from the typo.

## Testing

764 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -512,10 +512,13 @@ class PitBoss:
     async def set_wifi_update_frequency(self, fast=5, slow=60) -> dict | None:
         """Sets how often (in seconds) the device sends WiFi status updates.
 
+        The method name is `WiFi`, not `Wifi`. RPC names are matched exactly,
+        and every firmware image registers it with the capital F.
+
         :meta private:
         """
         return await self._conn.send_command(
-            "PB.SetWifiUpdateFrequency",
+            "PB.SetWiFiUpdateFrequency",
             await self._authenticate({"slow": slow, "fast": fast}),
         )
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -64,6 +64,7 @@ class FakeTransport(Transport):
     def __init__(self, password: str = ""):
         self.virtual_data: dict = {}
         self.fast_updates_requested = False
+        self.wifi_update_frequency: dict | None = None
         super().__init__()
         self.last_mcu_command: str | None = None
         self.password = password
@@ -93,6 +94,10 @@ class FakeTransport(Transport):
             "PB.GetTime": self._get_time,
             "PB.GetVirtualData": self._get_virtual_data,
             "PB.WiFiAwakeWDT": self._wifi_awake_wdt,
+            # Registered under the name the firmware uses -- `WiFi`, not
+            # `Wifi` -- so a wrong spelling falls through to the unknown
+            # method path rather than being quietly accepted.
+            "PB.SetWiFiUpdateFrequency": self._set_wifi_update_frequency,
             "PB.SetVirtualData": self._set_virtual_data,
             "PB.SetDevicePassword": self._set_password,
             "PB.SendMCUCommand": self._send_mcu_command,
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
One character, and the call has never once reached a grill. `fake_firmware/init.js:551` registers `PB.SetWiFiUpdateFrequency` with the capital F, so even the repo's own model of the firmware disagreed with the code — and `api.py:384`, in `request_fast_updates()`'s docstring, spells it correctly two hundred lines above the line that spells it wrong.

**The `RPC.Describe` pair is the cleanest evidence anyone has produced in this series.** Not "the docs say" or "the image contains", but the device answering `404 name not found` to one spelling and describing the other. That settles it in a way nothing else can.

## The blind spot is the part worth keeping

> The test asserted the name the code sent.

That is the third variant of the same failure in this repo: #551's mocks asserting `{}` that no grill returns, #546's docstring snippet that was never executed, and now a test whose expected value was copied from the implementation. All three pass forever while the real thing is broken, because the assertion and the code have the same author and the same mistake.

Driving it through `FakeTransport` instead is the right fix precisely because the fake **refuses a wrong spelling the way a grill does** — the assertion now has an independent source of truth. Confirmed by mutation: restoring the old spelling turns three tests red, and the description is right that a mocked `send_command` would still be green.

```
FAILED test_set_wifi_update_frequency[no_password]
FAILED test_set_wifi_update_frequency[with_password]
FAILED test_set_wifi_update_frequency_uses_the_name_the_firmware_registers
```

## Verified

* 764 tests, ruff, `ruff format --check`, mypy clean; CI green.
* `PB.WiFiAwakeWDT` in `request_fast_updates()` is a different RPC and unaffected, as you say — I checked it against `init.js:580` rather than assuming the two shared a spelling problem.
* Nothing downstream calls this: ha-pitboss does not reference `set_wifi_update_frequency` at all.

The method is `:meta private:` and unused, so the practical impact today is nil — but it is exactly the kind of thing that would cost someone an afternoon the first time they tried to use it. Labelled `bug`. Merging.

PR #560, opened 2026-08-03
Send login credentials in the request body

### LINKED ISSUE #539: auth: account credentials are sent in the URL query string
In `pytboss/auth.py`, `async_login()` does:

```python
params = {"email": email, "password": password}
async with session.post(f"{API_URL}/login/app", params=params) as response:
```

`params=` puts the account email and password into the URL query string, where they routinely end up in server access logs, proxy logs, and anything else that records request URLs. Moving them into the request body (`json=` or `data=`) would avoid that — but whether the Dansons endpoint accepts body-encoded credentials needs verifying against the live API, which is why this is an issue rather than a PR.

Smaller robustness items in the same function, happy to fold into the same fix once the transport question is settled:

- No `response.raise_for_status()` — a non-JSON error page surfaces as an opaque `ContentTypeError` from `response.json()`.
- `response.json()` is awaited twice (works because aiohttp caches the body, but the second call is redundant).
- `response_json["status"]` raises `KeyError` on any response that doesn't carry that field.

trailro:
Fixes #539.

`async_login()` sent the account email and password as URL query parameters, where they routinely end up in server/proxy access logs. As verified live in the issue discussion, the endpoint accepts the credentials as a JSON body with an identical success response (and the body takes precedence when both are present), so this moves them to `json=`.

While in there, this also fixes the error handling, which was dead code against the live API:

- Wrong credentials return HTTP 404 with `{"status": "error", ..., "errors": {"message": ...}}` — `errors` plural — so the old `response_json["error"]["message"]` raised `KeyError: 'error'` instead of `Unauthorized`. The function now parses `errors.message` and raises `Unauthorized` for any non-success status.
- Non-JSON responses (e.g. the bare-POST 302 to an HTML page) now surface via `raise_for_status()` or a clear `Unauthorized`, instead of an opaque `ContentTypeError`.
- Removed the redundant second `await response.json()`.

Tests now mock the live response shapes — the old error test mocked a singular-`error` shape, which is why the `KeyError` was never caught in CI.

Verified live against a real account: login via JSON body succeeds and returns the same token payload, and a wrong password now raises `Unauthorized("Customer not found or password entered incorrectly. Please try again.")`.

### DIFF
--- a/pytboss/auth.py
+++ b/pytboss/auth.py
@@ -1,6 +1,6 @@
 """Authentication routines."""
 
-from aiohttp import ClientSession
+from aiohttp import ClientSession, ContentTypeError
 
 from .exceptions import Unauthorized
 
@@ -17,12 +17,17 @@ async def async_login(
     :param password: The PitBoss account password.
     :raise pytboss.exceptions.Unauthorized: If the credentials are rejected.
     """
-    params = {"email": email, "password": password}
-    async with session.post(f"{API_URL}/login/app", params=params) as response:
-        response_json = await response.json()
-        if response_json["status"] == "error":
-            raise Unauthorized(response_json["error"]["message"])
-        token = (await response.json())["data"]["token"]
+    payload = {"email": email, "password": password}
+    async with session.post(f"{API_URL}/login/app", json=payload) as response:
+        try:
+            response_json = await response.json()
+        except ContentTypeError:
+            response.raise_for_status()
+            raise Unauthorized("Unexpected non-JSON response from login endpoint")
+        if response_json.get("status") != "success":
+            errors = response_json.get("errors") or {}
+            raise Unauthorized(errors.get("message", "Login failed"))
+        token = response_json["data"]["token"]
         return {
             "Accept": "application/json",
             "Content-Type": "application/json",
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1,14 +1,15 @@
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
+from aiohttp import ClientResponseError, ContentTypeError
 
 from pytboss.auth import API_URL, async_login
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
Credentials out of the query string is the right fix and the live check is what makes it safe to do in one move — that the body wins when both are present means there is no transition window where a stale caller sends both and gets the wrong one. @Giordano22 independently reaching the same conclusion in Postman is a useful second source for something this easy to get wrong once.

**The dead `Unauthorized` path is the better find of the two.** `response_json["error"]["message"]` against a live `errors` (plural) means a wrong password has been raising `KeyError: 'error'` rather than `Unauthorized` for as long as this has existed — and the test mocked the singular shape, so it passed while asserting a payload the API does not send. Same failure as #559's spelling: the expectation was copied from the implementation, so both were wrong together and agreed with each other.

Mutated the three behavioural changes separately:

```
revert json= to params=            -> test_async_login_success        FAILS
revert errors -> error             -> test_async_login_bad_credentials FAILS
revert != "success" to == "error"  -> 4 passed
```

**The third is worth naming.** For the two payloads under test the conditions are equivalent, so nothing distinguishes them — the inversion rests on your live login rather than on the suite. I think it is right regardless of what `status` turns out to hold for some third response: any non-success now raises a clear `Unauthorized` where the old code would have fallen through to `data.token` and died with a `KeyError`. So the change is an improvement whichever way an unknown status goes, and it is only *correct* rather than merely better because you confirmed a real login still succeeds through this code.

I cannot reproduce that — I have no account — so that is on your word, as is the 404-carrying-the-error-body detail that makes the ordering (`json()` first, `raise_for_status()` reserved for non-JSON) necessary rather than stylistic.

753 tests, ruff, `ruff format --check` and mypy clean; CI green. Labelled `bug`. Merging.

PR #561, opened 2026-08-03
Wrap RPC.List and PBL.GetLoaderVersion

trailro:
Two of the calls from #167, picked because they are reads that take no arguments, need no password, and cannot change anything. Rationale for the other thirteen is [in the issue](https://github.com/dknowles2/pytboss/issues/167).

## `list_rpcs()`

What a grill answers is not uniform, and the library has had no way to ask. Both #543 and #458 currently turn on that question — which grills serve a local endpoint, which serve which `OTA.*` method — and the answer has been reached by inference rather than by asking. This is the call that asks.

A PB1600PS1 on 0.5.7 lists 56 methods, and that list is where `PBL.GetLoaderVersion`, the `Wifi.*` calls, and the `OTA.*` set do or do not appear.

Not universal itself, and the docstring says so: the ESP-IDF line names it `RPC.ListEx` and does not serve `RPC.List`. I did not add a fallback between them — which to call is a question about the grill, and inventing a two-call retry here would bury that rather than answer it.

## `get_loader_version()`

The loader application carries the grill's own application and is versioned separately: the same grill that reports firmware `0.5.7` reports loader `0.2.2`. Present on both firmware lines.

## Verified

Both against a PB1600PS1 on 0.5.7, read-only:

```
GET /rpc/PBL.GetLoaderVersion  -> {"loaderVersion":"0.2.2"}
GET /rpc/RPC.List              -> ["PBL.GetWifiScanStatus","PBL.StartWifiScan", ... ]  (56)
RPC.Describe both              -> args_fmt ""   (no parameters)
```

The fake transport answers with those shapes — a bare **list** rather than an object for `RPC.List`, which is why `list_rpcs()` returns `list[str]` and guards the type rather than assuming a dict.

## Testing

770 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -585,3 +585,35 @@ class PitBoss:
         :param timeout: Time (in seconds) after which to abandon the RPC.
         """
         return await self._conn.send_command("RPC.Ping", {}, timeout=timeout)
+
+    async def list_rpcs(self) -> list[str]:
+        """Returns the RPC methods this grill serves.
+
+        What a grill answers is not uniform, so this is how a caller finds
+        out rather than inferring it from a model or firmware version. A
+        PB1600PS1 on 0.5.7 lists 56 methods; the same list is where
+        `PBL.GetLoaderVersion` and the `Wifi.*` calls do or do not appear.
+
+        Not universal itself: the ESP-IDF firmware line -- versioned `16.x`,
+        on the PBC2, PBD, PBE, PBL2 and PBT boards -- names this one
+        `RPC.ListEx` and does not serve `RPC.List` at all, so a caller
+        covering both has to ask for each.
+        """
+        result = await self._conn.send_command("RPC.List", {})
+        return result if isinstance(result, list) else []
+
+    async def get_loader_version(self) -> str | None:
+        """Returns the version of the loader application, if it has one.
+
+        This is the `PitBoss Loader` that carries the grill's own
+        application, and is versioned independently of the firmware
+        `get_firmware_version()` reports: a grill on firmware 0.5.7 answers
+        `0.2.2` here.
+
+        `None` when the grill does not serve `PBL.GetLoaderVersion`, which
+        `list_rpcs()` reports without a round trip that errors.
+        """
+        result = await self._conn.send_command("PBL.GetLoaderVersion", {})
+        if isinstance(result, dict):
+            return result.get("loaderVersion")
+        return None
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Both are good picks for the same reason: reads, no arguments, no password, nothing to get wrong. And `list_rpcs()` in particular is the call that turns a recurring inference in this repo into a question — #543 and #458 both hinge on what a given grill serves, and until now that has been argued from firmware images rather than asked.

Not adding a `RPC.List` / `RPC.ListEx` fallback is right. Which one a grill answers *is* the fact a caller wants; a silent two-call retry would hide it behind a result that looks uniform.

## The bare-list return exposes something #551 missed, and I want it recorded

`RPC.List` answers with a JSON **array**. But `send_command` is annotated `dict[Any, Any] | None` — narrowed there four hours ago on an audit that established "every getter answers with an object". That audit covered the `PB.*` handlers the app registers; **`RPC.List` is a Mongoose core RPC, so it was not in `fake_firmware` to be audited.** mypy agrees the declaration is what I say it is:

```
Revealed type is "dict[Any, Any] | None"
```

So the annotation is still not the whole truth, and your `isinstance(result, list)` guard is doing real work at runtime while being **invisible to the type checker** — mypy accepts the true branch only because it believes it unreachable. I checked what that costs:

```
guard removed entirely      -> mypy: Success        (it cannot see the guard mattering)
bare `return result`        -> mypy: 1 error        (it does catch the naive simplification)
```

So the important failure is caught and the guard is not load-bearing against mypy — but nothing would stop someone deleting it as redundant. Worth its own issue against `send_command`'s annotation rather than widening it here; I filed the original as #542 and would rather not reopen that thread inside an unrelated PR. **The lesson for next time is that an audit over `fake_firmware` covers the app's handlers and not Mongoose's own** — the same blind spot that left `OTA.*`, `Sys.*` and `FS.*` unmodelled.

## Verified

* 770 tests, ruff, `ruff format --check`, mypy clean; CI green.
* The fake answering `RPC.List` with a bare list rather than an object is what makes the test meaningful — a dict-shaped fake would have let the guard pass vacuously, which is the #559 failure again.
* `get_loader_version()` returning `None` rather than raising when the method is absent matches how the rest of the library treats "this grill does not do that".

The loader being versioned separately — `0.2.2` under firmware `0.5.7` — is worth having in a docstring; it is exactly the sort of thing that reads like a bug in a diagnostics panel otherwise.

The ESP-IDF caveat rests on the same static analysis as #558 and I can corroborate none of it. Labelled `enhancement`. Merging.

PR #563, opened 2026-08-03
Bound the BLE RPC response read

trailro:
`_on_rpc_data_received` takes the reply length from the control notification and reads until it has that many bytes, with no cap, no progress check, and the lock `_send_prepared_command` needs held throughout.

```python
resp_len = _decode_len(data)
resp = bytearray()
async with self._lock:
    while len(resp) < resp_len:
        resp += await self._ble_client.read_gatt_char(CHAR_RPC_DATA)
```

## Two ways out of that loop, and neither is "finish"

**A read that returns nothing.** The loop spins on it, and nothing in that path suspends — `read_gatt_char` returns without yielding once there is nothing to wait for. Driven with a client whose reads go empty:

```
read_gatt_char calls: 200001 in 0.03s
other tasks that got to run meanwhile: 0
```

Zero. On a Home Assistant instance that is not this integration hanging, it is the whole event loop.

**A length that is merely implausible.** `_decode_len` reads four bytes, so a corrupt or desynchronised notification can ask for up to 4 GiB. At ~20 bytes and ~30 ms per GATT read that is hours of real round trips for a reply that will never arrive — all of it holding the lock, so every command sent meanwhile waits out its own 30 s timeout behind it.

## The change

Reject a length past anything a reply can be, and stop when a read stops making progress. Both abandon the reply and log, which leaves the caller to time out exactly as it already does whenever a reply goes missing — no new failure mode, just a bounded one.

The cap is 64 kB. For scale, the largest replies the library can provoke are `FS.Get`, capped at 512 bytes of content, and `RPC.List`, about 1.4 kB on a grill serving 56 methods.

The lock is deliberately kept: it serialises access to the same characteristic `_send_prepared_command` writes to, so the fix is to bound how long it is held rather than to release it.

## Testing

Two tests. The empty-read one **fails on the unpatched code** — the old loop reads straight past the empty chunk and exhausts the fake's replies. The implausible-length one asserts no read is even attempted; on the unpatched code that path is the defect itself, so it is written to assert the absence rather than to wait for a loop that does not end.

778 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -40,6 +40,14 @@ CHAR_RPC_RX_CTL = _uuid("_mOS_RPC_rx_ctl_")
 DisconnectCallback = Callable[[BleakClient], None]
 """A callback function called when the BLE connection is disconnected."""
 
+MAX_RPC_RESPONSE_SIZE = 64 * 1024
+"""Largest reply this transport will try to assemble, in bytes.
+
+The length comes off the wire, so a corrupt or desynchronised control
+notification can ask for up to 4 GiB. Nothing the library sends provokes a
+reply anywhere near this bound: the largest are `FS.Get`, capped at 512 bytes
+of content, and `RPC.List`, about 1.4 kB on a grill serving 56 methods."""
+
 
 class BleConnection(Transport):
     """Bluetooth LE protocol transport for Mongoose OS devices."""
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
**I reproduced the starvation on unpatched `main` and got your numbers exactly.** A client whose reads go empty, with an unrelated task on the loop trying to tick:

```
read_gatt_char calls: 200001 in 0.02s
other tasks that got to run meanwhile: 0
```

Zero. Not "starved" as a manner of speaking — the loop never yields at all, because nothing in that path suspends once there is nothing to wait for. Inside Home Assistant that is every integration on the instance, not this one hanging. On the branch the same probe reads once and returns.

That is the most serious thing found in either repo today, and it is reachable from a truncated or desynchronised notification rather than anything exotic.

**Both guards are pinned, and the second one is pinned in an unusual way:**

```
remove the empty-chunk guard -> test_on_rpc_data_received_gives_up_on_an_empty_read  FAILS
remove the size cap         -> the suite HANGS
```

The cap's test times out rather than failing, because without the cap that test *is* the defect — it walks into the unbounded loop. You said as much in the description, and writing it to assert that no read is attempted is the right way round: waiting for a loop that does not end is not a test.

**The cap is well chosen and the reasoning is checkable.** I sized `RPC.List` independently — 56 method names is about 1.2 kB of JSON, so 64 kB is roughly 50x the largest reply the library can provoke, and `FS.Get` is capped at 512 bytes of content by its own chunking. Wide enough that nothing legitimate is near it, narrow enough that the bad case is bounded in milliseconds rather than hours.

**Keeping the lock rather than releasing it is right**, and worth that you said why: it serialises the characteristic `_send_prepared_command` writes to, so releasing it would trade a bounded stall for a data race. Bounding how long it is held is the fix.

Abandoning the reply and letting the caller time out is also the right failure: it is the path that already exists for a reply that never arrives, so nothing downstream learns a new behaviour.

778 tests, ruff, `ruff format --check` and mypy clean; CI green. Labelled `bug`. Merging.

PR #564, opened 2026-08-03
Deliver virtual data over the WebSocket transport

trailro:
`subscribe_vdata()` has never fired on a WebSocket connection.

## What the firmware sends

`sendWSStatus` builds one object and attaches virtual data to it, rather than sending a frame of its own:

```js
let data = { id: -1, src: deviceId, status: wsStatus };
if (sendVData === true && vData !== null) { data.data = vData; }
if (pState !== "") { data.pState = pState; }
WS.send(wsConn, JSON.stringify(data));
```

So virtual data arrives as `payload["data"]`, on the same frame as `status`. `_handle_message` matched `"status"` first and returned, dropping it every time.

(That is from the vendor image for 0.5.7 rather than from `fake_firmware/` — same code, and I wanted the shape from the thing that actually ships. It is identical in every mJS version from 0.2.3 to 0.6.0.)

## The branch that was supposed to catch it

```python
if payload.get("result", None):
    # TODO: Verify this is actually a vdata response.
```

Unreachable: every RPC reply carries an `id` and is claimed by the branch above it, and the firmware sends no other frame shaped like that. The TODO asked whether this really is a vdata response — it is not, so the branch is removed rather than repaired. Happy to keep it as a fallback if you would rather; I could not construct a payload that reaches it.

## The type mismatch behind it

`PitBoss._on_vdata_received` does `json.loads(payload)` unconditionally. That is right for `ble`, which reads virtual data off the debug log while it is still the text the firmware printed, and wrong for `wss`, where it is a member of an already-parsed frame. Had the old branch ever fired it would have raised `TypeError: the JSON object must be str, bytes or bytearray, not dict`. Both shapes are now accepted, and `RawVDataCallback` says so.

## Testing

The old tests exercised `{"result": "some-vdata"}` — a shape the firmware never sends — so they passed against a transport that delivered nothing. Replaced with the real frame: virtual data and state both dispatched from one push, a push without `data`, a push whose `data` has no subscriber, and an api-level test that both the decoded and the text form reach subscribers.

778 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

## Not verified on hardware

I have a grill on the local network and over BLE, but not on the Dansons relay — the cloud path has never worked here, which is what #505 is about. The frame shape is read from the firmware that builds it, not observed on the wire.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -194,8 +194,12 @@ class PitBoss:
             for callback in callbacks:
                 await _invoke(callback, self._state)
 
-    async def _on_vdata_received(self, payload: str):
-        vdata = json.loads(payload)
+    async def _on_vdata_received(self, payload: str | dict):
+        # Both, because the transports differ in what they can hand over.
+        # `ble` reads virtual data off the debug log, where it is still the
+        # text the firmware printed; `wss` receives it as a member of an
+        # already-parsed status frame, so it arrives decoded.
+        vdata = json.loads(payload) if isinstance(payload, str) else payload
         _LOGGER.debug("VData received: %s", vdata)
         async with self._lock:
             callbacks = list(self._vdata_callbacks)
--- a/pytboss/transport.py
+++ b/pytboss/transport.py
@@ -28,7 +28,7 @@ class RawStateCallback(Protocol):
     ) -> None: ...
 
 
-RawVDataCallback = Callable[[str], Awaitable[None]]
+RawVDataCallback = Callable[[str | dict], Awaitable[None]]
 """Callback invoked by a `Transport` with a raw, unparsed VData payload."""
 
 RPCResult = dict[Any, Any] | list[Any] | None
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -196,6 +196,16 @@ class WebSocketConnection(Transport):
             return
 
         if "status" in payload:
+            # Virtual data rides on the status push rather than arriving on a
+            # frame of its own. The firmware builds one object and attaches it
+            # only when there is something to send::
+            #
+            #     let data = {id: -1, src: deviceId, status: wsStatus};
+            #     if (sendVData === true && vData !== null) { data.data = vData; }
+            #
+            # so it has to be picked up here, before returning.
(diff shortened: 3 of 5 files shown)


dknowles2 (approved):
**The repo's own model corroborates the vendor image you read, line for line.** `fake_firmware/init.js:189-198`:

```js
let data = { id: -1, src: deviceId, status: wsStatus };
if (sendVData === true && vData !== null) { data.data = vData; }
```

So virtual data really does ride on `payload["data"]` alongside `status`, and `_handle_message` matching `"status"` first and returning drops it every time. Nice that the two sources agree — this is one of the few claims here where I could check both.

**And the `id: -1` on that frame is what settles the reachability of the branch you removed.** The status push carries both `status` and `id`, so it is claimed by the first branch; every RPC reply carries its own `id` and is claimed by the second. Nothing the firmware sends can reach `payload.get("result")` at all. Your "I could not construct a payload that reaches it" is not just an absence of imagination — the branch ordering above it makes it unreachable by construction. Removing it rather than repairing it is right, and I would not keep it as a fallback.

The `TODO: Verify this is actually a vdata response` is the sort of comment that turns out to be a question nobody answered for years. The answer was no.

**The type mismatch is the good catch behind the good catch.** `json.loads(payload)` unconditionally would have raised `TypeError: the JSON object must be str, bytes or bytearray, not dict` the first time that branch ever fired — so even a payload that *did* reach it would have died. Two bugs stacked, and the outer one hid the inner one. Accepting both shapes and saying why in the comment — `ble` reads it as printed text, `wss` as a member of a parsed frame — is the honest fix rather than making one transport pretend to be the other.

**Dispatching vdata before the `if not self._state_callback: return`** is a detail worth not getting wrong: a consumer subscribed only to vdata still receives it.

## Verified

* Mutated the dispatch away — `test_vdata_arrives_on_the_status_push` fails and nothing else.
* The old tests exercised `{"result": "some-vdata"}`, a shape the firmware never sends. That is the same failure as #559's spelling and #560's singular `error`: an expectation written from the implementation, passing forever against a transport that delivered nothing. Replacing them with the real frame is what makes this stick.
* 778 tests, ruff, `ruff format --check` and mypy clean; CI green.

Labelled `bug`. Merging.

PR #565, opened 2026-08-03
Release the BLE client when the link goes, or never opens

trailro:
Two places keep a `BleakClient` that nothing else will clean up.

## An unsolicited disconnect

`_on_disconnected` marks the transport disconnected but keeps the reference, and `_send_prepared_command` decides whether it can send from that reference alone. So a grill switched off or carried out of range turns the next command into a raw `BleakError`, where an explicit `disconnect()` gives `NotConnectedError` — `_disconnect_locked` already clears it, with a comment saying why.

```
OLD: _ble_client still set? True
OLD raised: BleakError: Not connected
```

Worth noting how that probe has to be written, because it fooled me first time round: `_on_disconnected` schedules `_fail_pending_commands()`, and if the next command is issued before that task runs, its future is resolved by the cleanup and the call *does* raise `NotConnectedError` — for an unrelated reason. Letting the loop settle first is what shows the real behaviour, and the test does that.

The clear is guarded on identity: a reconnect may already have put a live client there, and a late callback for the previous one must not drop it.

## A connection that never finishes opening

`_connect_locked` left a **connected** client behind if `start_notify` raised. The connection itself had succeeded, so nothing disconnected it, and the next `connect()` overwrote the only reference to it. Each retry burned a slot on an adapter that has very few.

```
after failed connect, _ble_client is c1 -> True | c1.disconnect called -> False
after retry,          _ble_client is c2 -> True | c1 ever disconnected -> False
```

The existing test asserts `is_connected()` is False afterwards, which was already true; it never checked that the client was released.

## Testing

Three tests, two of them **verified to fail on the unpatched code** (restored `ble.py` from `main`, kept the tests). The third — a disconnect callback for a client that has since been replaced — guards the identity check rather than a defect.

783 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -110,14 +110,37 @@ class BleConnection(Transport):
         # Only once the notifications are registered: until they are there is
         # no receive path, so a command would be sent and never answered while
         # `is_connected()` claimed otherwise.
-        await self._ble_client.start_notify(CHAR_RPC_RX_CTL, self._on_rpc_data_received)
-        await self._ble_client.start_notify(CHAR_DEBUG_LOG, self._on_debug_log_received)
+        try:
+            await self._ble_client.start_notify(
+                CHAR_RPC_RX_CTL, self._on_rpc_data_received
+            )
+            await self._ble_client.start_notify(
+                CHAR_DEBUG_LOG, self._on_debug_log_received
+            )
+        except Exception:
+            # The connection itself succeeded, so leaving it behind holds a
+            # slot on an adapter that has few, and the next `connect()` would
+            # overwrite the only reference to it.
+            client, self._ble_client = self._ble_client, None
+            try:
+                await client.disconnect()
+            except Exception as ex:  # noqa: BLE001
+                _LOGGER.debug("Failed to release the client: %s", ex)
+            raise
         self._is_connected = True
 
     def _on_disconnected(self, client: BleakClient) -> None:
         """Called when our Bluetooth client is disconnected."""
         _LOGGER.debug("Bluetooth disconnected.")
         self._is_connected = False
+        # Released here as well as in `_disconnect_locked`, which drops it for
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Reproduced both, and the first exactly as you wrote it:

```
main:    _ble_client still set? True    raised: BleakError: Not connected
branch:  _ble_client still set? False   raised: NotConnectedError
```

**And the warning about how to write that probe is the most useful thing in the description.** I hit it too — my first version sent the command immediately after `_on_disconnected` and got `NotConnectedError`, which looked like the bug was not there. It was the scheduled `_fail_pending_commands()` resolving the future, not the client reference doing its job. Letting the loop settle first is what separates the two, and a reviewer who did not know that would have concluded there was nothing to fix.

That is a genuinely adversarial detail to have volunteered, since it would have made your own PR look unnecessary.

**The identity guard is right and not obvious.** `if self._ble_client is client` rather than an unconditional clear, because a reconnect may have already installed a live client by the time a late callback for the previous one arrives — dropping that would be a worse bug than the one being fixed.

**The `start_notify` case is the one with a resource cost.** The connection succeeded, so nothing disconnected it, and the next `connect()` overwrote the only reference — each retry burning a slot on an adapter that has very few. Swapping the reference out *before* awaiting `disconnect()` is the correct order: it cannot be handed out again while the release is in flight.

And the point about the existing test is exactly right — asserting `is_connected()` is False afterwards passed already, because that was never the part that was broken. It checked the flag and not the thing the flag is about.

## Verified

```
keep the client on unsolicited drop         -> test_an_unsolicited_disconnect_releases_the_client  FAILS
leave a connected client on notify failure  -> test_a_failed_start_notify_releases_the_connection  FAILS
```

Each fix has exactly one test, and neither passes without it. 783 tests, ruff, `ruff format --check` and mypy clean; CI green.

Labelled `bug`. Merging.

PR #566, opened 2026-08-03
Survive replies the grill really sends

trailro:
Two places assume a reply carries a field it may not.

## `get_state()` indexed both frames

```python
status = self.spec.control_board.parse_status(resp["sc_11"]) or {}
status.update(self.spec.control_board.parse_temperatures(resp["sc_12"]) or {})
```

The firmware blanks both the moment it forwards a command to the MCU:

```js
function sendMCUCommand(pCommand) {
  lastStatus.sc_11 = "";
  lastStatus.sc_12 = "";
  uartSend(mcuCommand);
}
```

and refills them from the next reply, so a poll landing in that window is answered with empty strings — which the old code handled, since `parse_status("")` returns `None`. A reply *missing* either key did not:

```
get_state no sc_ keys  -> KeyError: 'sc_11'
get_state only sc_11   -> KeyError: 'sc_12'
get_state None reply   -> KeyError: 'sc_11'
```

Neither frame is indexed now, and either may be absent or empty.

## `get_uptime()` cached a reply it could not read

`result.get("time", 0.0)` turned an unreadable reply into the cache, and with it every `timed_key` for the full 60-second TTL. On a password-protected grill that is a minute of rejected commands from one bad reply, and the next call does not go back to the grill because the cache looks fresh:

```
get_uptime malformed -> 0.0; cached=0.0
next get_uptime      -> 0.00018   (should have re-read 500.0)
```

It now leaves the cache alone and answers from what it already had, so the next call retries.

## Testing

Five tests. **Four fail on the unpatched code** — the fifth is the blanked-frames case, which already worked and is pinned so it keeps working.

785 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -490,8 +490,20 @@ class PitBoss:
         resp = as_dict(
             await self._conn.send_command("PB.GetState", await self._authenticate({}))
         )
-        status = self.spec.control_board.parse_status(resp["sc_11"]) or {}
-        status.update(self.spec.control_board.parse_temperatures(resp["sc_12"]) or {})
+        # Either frame can be absent or empty, so neither is indexed. The
+        # firmware blanks both the moment it forwards a command to the MCU::
+        #
+        #     function sendMCUCommand(pCommand) {
+        #       lastStatus.sc_11 = "";
+        #       lastStatus.sc_12 = "";
+        #
+        # and refills them from the next reply, so a poll landing in that
+        # window is answered with empty strings rather than an error.
+        status = StateDict()
+        if sc_11 := resp.get("sc_11"):
+            status.update(self.spec.control_board.parse_status(sc_11) or {})
+        if sc_12 := resp.get("sc_12"):
+            status.update(self.spec.control_board.parse_temperatures(sc_12) or {})
         if status:
             async with self._lock:
                 self._state.update(status)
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
All three reproduce on `main` exactly as listed:

```
get_state no sc_ keys  -> KeyError: 'sc_11'
get_state only sc_11   -> KeyError: 'sc_12'
get_state None reply   -> KeyError: 'sc_11'
```

And the distinction you draw is the precise one: `parse_status("")` returns `None`, so the **empty**-string case was already handled — what was not was a key being **absent**. Those look alike and are not, and it would be easy to read the firmware blanking the frames and conclude the code already coped.

**The firmware quote checks out against the repo's own model**, `fake_firmware/init.js:315-321`:

```js
function sendMCUCommand(pCommand) {
  let mcuCommand = fromHexStr(pCommand);
  lastStatus.sc_11 = "";
  lastStatus.sc_12 = "";
```

So every MCU command opens a window where a poll is answered with blanks — which is frequent rather than exotic, especially at #351's ten-second cadence while the grill is running.

**The `get_uptime` half is the better change and the reasoning is what makes it so.** Caching a missing `time` as `0.0` would have been the obvious fix and is the harmful one: `timed_key` would be built from a wrong uptime for the whole 60-second TTL, so a single unreadable reply costs a minute of rejected commands on a password-protected grill. Extrapolating from the last good value instead — and only falling back to `0.0` when there has never been one — keeps the failure to the one reply rather than the window.

That is the same asymmetry #538 turned on, and it is good to see it applied rather than restated.

`isinstance(uptime, (int, float))` rather than a truthiness check also matters here: a genuine `0.0` uptime, from a grill that has just restarted, is exactly when the cache most needs to be right.

785 tests, ruff, `ruff format --check` and mypy clean; CI green. Labelled `bug`. Merging.

PR #568, opened 2026-08-03
Read a file by what the device sends, not by what was asked for

trailro:
`get_file_content` advances the offset by the 512 bytes it asked for, however many actually came back, and indexes `resp["data"]` and `resp["left"]` without checking either is there.

```python
content += b64decode(resp["data"])
offset += length
if resp["left"] == 0:
```

## Three ways that goes wrong

**A device that answers with less than was asked for leaves a hole, and the read reports success.** The reply is bounded by the RPC frame size, which is configurable — `Config.Get {key: "rpc"}` reports `max_frame_size: 4096` on the grill I have, and a smaller one is a supported setting. Reading a 1260-byte file in 200-byte answers:

```
asked for 1260 bytes, got back 600
identical to the file: False
```

No exception. The caller gets a file with the middle missing and no way to know.

**A `left` that never reaches zero asks for the same offset forever.** I stopped the probe at 300 round trips; nothing in the loop would have ended it.

**A reply without `data` raised `KeyError: 'data'`** — the same shape #566 has just been through elsewhere in the library.

## The change

The offset follows what arrived, and the loop stops when nothing does. That last part ends the runaway case as well, since reading past the end of a file returns nothing whatever `left` claims.

## Testing

Three tests. Two **fail on the unpatched code**; the third — the device that stops sending — is the runaway, so on the old code it does not fail, it hangs. I checked that separately rather than leaving it in a suite run.

791 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/fs.py
+++ b/pytboss/fs.py
@@ -36,12 +36,23 @@ class FileSystem:
                     "FS.Get", {"filename": filename, "offset": offset, "len": length}
                 )
             )
-            content += b64decode(resp["data"])
-            offset += length
-            if resp["left"] == 0:
-                # Decoded once, whole: a multi-byte character split across
-                # the chunk boundary is not valid UTF-8 on its own.
-                return content.decode("utf-8")
+            chunk = b64decode(resp.get("data", ""))
+            if not chunk:
+                # Nothing came back, so nothing more will: reading on would
+                # ask for the same offset forever when `left` disagrees.
+                break
+            content += chunk
+            # Advanced by what arrived rather than by what was asked for. A
+            # device answering with less -- the reply is bounded by the RPC
+            # frame size, which is configurable and smaller than this on some
+            # units -- would otherwise leave a hole in the middle of the file
+            # and the read would report success.
+            offset += len(chunk)
+            if resp.get("left") == 0:
+                break
+        # Decoded once, whole: a multi-byte character split across a chunk
+        # boundary is not valid UTF-8 on its own.
+        return content.decode("utf-8")
 
     async def set_file_content(
         self, filename: str, data: str | bytes, append: bool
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Reproduced both, and the first is the one that would be hardest to ever diagnose:

```
main:      asked for 1260 bytes, got back 600   identical to the file: False
branch:    asked for 1260 bytes, got back 1260  identical to the file: True
```

**A file with the middle missing, returned as a success.** No exception, no short-read indication, nothing in the value that says it is incomplete — and since `get_file_content` decodes to `str`, a caller gets something that looks exactly like a file. Every other bug in this pair of repos announces itself eventually; this one hands you plausible wrong data and stops.

The runaway too:

```
main:      still going at 301 round trips
branch:    returned after 1
```

**And the reachability argument is the part I would have wanted and did not expect to get.** "The device might answer short" is a hypothesis; `max_frame_size: 4096` from `Config.Get {key: "rpc"}` on a real grill, and it being a supported setting that can be smaller, is a reason. That turns this from hardening into a bug with a route to it.

**Advancing the offset by what arrived rather than by what was asked for is the fix**, and it is the kind of line that looks like a tidy-up until you see what the old one does. Stopping on an empty chunk closing the runaway as a side effect is a genuine simplification rather than two guards bolted together — reading past the end returns nothing whatever `left` claims, so the same condition covers both.

`resp.get("data", "")` and `resp.get("left")` are the same lesson as #566, applied without being asked. That is three places now where a reply was indexed for a key it might not carry.

**Saying that the third test hangs rather than fails on old code, and that you checked it separately, is the right disclosure.** A suite run that never finishes is not a failing test, and someone re-running your verification would otherwise think they had broken something.

791 tests, ruff, `ruff format --check` and mypy clean; CI green. Labelled `bug`. Merging.

PR #569, opened 2026-08-03
Make disconnect() final for the HTTP transport

trailro:
Targets `http-transport`, same as #546 and #558 — merge it into #543 with one click, take the diff, or ignore it.

**One of three**, from reviewing this PR's code. Suggested order: this one, then the fire-and-forget fix, then the contract items. They touch the same small file, so whichever lands first will leave the others needing a rebase — say the word and I will do it as they go.

## The defect

`_send_prepared_command` decides whether it can send from the session alone:

```python
if self._session is None or self._session.closed:
    raise NotConnectedError("Not connected")
```

and a caller's session is not this transport's to close — `_close_session()` says so itself. So after `disconnect()` the session is still open, commands still go out and succeed, and line 178's `self._connected = True` puts the transport back to reporting itself connected after it was told to stop.

Home Assistant always supplies a shared session (`async_get_clientsession`), so this is the ordinary case rather than an exotic one: a coordinator refresh still in flight when the entry unloads keeps polling the grill, and `is_connected()` flips back to `True` behind it. The same path lets a command go out before `connect()` has ever run, which is the thing that checks the grill is there at all.

## Why a separate flag rather than gating on `_connected`

Because the two answer different questions, and this class deliberately recovers from the second. A request that fails leaves `_connected` false, and the next call is still expected to work — the class docstring is explicit that there is no reconnect loop because *"a grill that goes away surfaces as `NotConnectedError` on the next call and starts working again by itself"*. Gating sends on `_connected` would turn one failed request into a permanently dead transport.

`_started` is therefore only about `connect()`/`disconnect()`, and `_connected` keeps meaning what its docstring says.

## Consistency

`wss.py` re-reads `_sock` under the lock and raises `NotConnectedError`; `ble.py` releases `_ble_client` so the same check refuses (#565). This was the third transport and the only one that did not.

## Testing

Three tests. Two **fail on this PR as it stands** — a command after `disconnect()` and a command before `connect()` both reach the grill. The third pins the recovery behaviour so the fix cannot quietly take it away: a failed request followed by a good one still works.

785 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -98,6 +98,10 @@ class HttpConnection(Transport):
         # ours to close, and it is rebuilt by `connect()` so this transport is
         # reusable after `disconnect()`.
         self._session_owned = session is None
+        # Two different questions. `_started` is whether this transport has
+        # been connected and not since disconnected; `_connected` is whether
+        # the grill answered the last time we spoke to it.
+        self._started = False
         self._connected = False
 
     async def connect(self) -> None:
@@ -115,17 +119,25 @@ class HttpConnection(Transport):
                 raise NotConnectedError("The session given to this transport is closed")
             self._session = ClientSession(loop=self._loop)
         # Set before the ping so `_send_prepared_command` will proceed; a
-        # failure below clears it again.
+        # failure below clears them again.
+        self._started = True
         self._connected = True
         try:
             await self.send_command("RPC.Ping", {})
         except Exception:
+            self._started = False
             self._connected = False
             await self._close_session()
             raise
 
     async def disconnect(self) -> None:
-        """Releases the session, if this transport owns one."""
+        """Stops using the grill, and releases the session if this owns one.
+
+        Final until `connect()` is called again. A caller's session stays
+        open, since it is theirs, so nothing else here would stop a command
+        from going out on it.
+        """
+        self._started = False
         self._connected = False
         await self._close_session()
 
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Reproduced, and it is worse than "commands still go out" — the transport puts itself back to reporting connected:

```
after disconnect(): command SUCCEEDED, is_connected=True
```

My bug, and the reasoning that produced it is visible in my own code: `_close_session()` refuses to close a caller's session, correctly, and I then wrote the send guard as if the session were the only thing that could stop a send. For an owned session the two coincide, which is why every test I wrote passed.

**And Home Assistant always supplies a shared session**, so this is the ordinary path rather than a corner: `async_get_clientsession` means the guard never bites there at all. A refresh in flight when the entry unloads keeps polling the grill, and `is_connected()` flips back behind it.

**The argument for a separate flag is the right one and I would have got this wrong too.** Gating sends on `_connected` looks like the tidy fix — one flag instead of two — and it would contradict the class docstring I wrote three paragraphs above it: no reconnect loop, because a grill that goes away *"starts working again by itself"*. One failed request would become a permanently dead transport. The two flags answer different questions and now say so.

Setting `_started` before the ping and clearing it on failure keeps `connect()`'s own send working, which is the detail that makes the flag actually placeable.

CI does not run here — pytboss only runs it for PRs targeting `main` — so the local run is the check: **794 tests, ruff and mypy clean.**

Merging this first, as you suggest. I will take the other two in order and rebase whichever needs it rather than asking you to.

dknowles2:
Correction to my review above: I wrote **794 tests**; the run was **785**. Ruff and mypy clean as stated, and the reproduction stands — I mistyped the number rather than the result, but a figure quoted as evidence should be the one that came out of the terminal.

Holding you to executed counts all day and then not checking my own would be the wrong way round.

PR #570, opened 2026-08-03
Send a no-answer command without waiting for one

trailro:
Targets `http-transport`, same as #546, #558 and #569.

**Two of three**, from reviewing this PR's code. It sits after #569 in the order I would merge them; both touch `disconnect()`, so whichever lands first leaves the other needing a rebase — happy to do it as they go.

## The defect

`send_command_without_answer` is inherited, and the base implementation awaits `_send_prepared_command`. On the other transports that is the whole of the send — `ble.py` writes to the characteristic, `wss.py` sends on the socket, both return immediately. Over HTTP a request is not finished until its response has been read.

So the one method whose entire purpose is "do not wait for a reply" waits for one. Its callers pick it for commands the board cannot answer:

```python
async def reboot(self) -> None:
    """...No response is expected, since the board reboots before it can answer."""
    await self._conn.send_command_without_answer("Sys.Reboot", {})
```

Over this transport `reboot()` and `Config.save_config(reboot=True)` block for the constructor timeout — 30 seconds by default — and then raise `NotConnectedError` at a caller with no reason to catch anything. Measured with a 1-second timeout to keep it short:

```
PROBE reboot: raised NotConnectedError after 1.00s
```

## The change

The request is issued in the background. The task is held in a set so it is not collected mid-flight, discarded on completion, and cancelled by `disconnect()` if one is still going.

`timeout` becomes meaningless on this path and the docstring says so, rather than accepting it and quietly ignoring it.

I considered keeping it awaited with a short timeout and swallowing the expiry, but that trades a 30-second block for a shorter one and still guesses how long a board takes to stop answering. Say the word if you would rather have that shape.

## Testing

Two tests, both **failing on this PR as it stands**: the call returns promptly against a server that never answers (while the request does reach it), and `disconnect()` clears a request still in flight.

784 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -43,6 +43,7 @@ it needs an answer for every grill rather than a raise for some.
 See https://github.com/dknowles2/pytboss/issues/505.
 """
 
+import asyncio
 import logging
 from asyncio import AbstractEventLoop
 from typing import Any
@@ -103,6 +104,7 @@ class HttpConnection(Transport):
         # the grill answered the last time we spoke to it.
         self._started = False
         self._connected = False
+        self._pending: set[asyncio.Task] = set()
 
     async def connect(self) -> None:
         """Verifies the grill answers RPC at this address.
@@ -139,6 +141,9 @@ class HttpConnection(Transport):
         """
         self._started = False
         self._connected = False
+        for task in list(self._pending):
+            task.cancel()
+        self._pending.clear()
         await self._close_session()
 
     async def _close_session(self) -> None:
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Reproduced against the branch as it stood, with a 1-second timeout to keep it short, and confirmed the fix:

```
before:  reboot: raised NotConnectedError after 1.00s
after:   reboot: returned after 0.00s
```

So the one method whose entire purpose is not waiting for a reply waited for one, then raised at a caller with no reason to catch anything. With the constructor default that is **30 seconds** for `reboot()` and for `Config.save_config(reboot=True)`.

My bug, and the shape of it is worth naming: I inherited `send_command_without_answer` because on `ble` and `wss` the base implementation is already correct — writing to a characteristic or a socket *is* the whole send. Over HTTP a request is not finished until its response is read, so the same base method means the opposite thing. Inheriting something that is right for every existing subclass is exactly how you end up not looking at it.

**Holding the task and cancelling it in `disconnect()` is the part I would have left out.** A bare `create_task` would have satisfied the probe above and left a task the collector can drop mid-flight, plus one still running after the transport was told to stop. `add_done_callback(self._pending.discard)` keeps the set from growing, which is the same care as #347's timer handle.

Swallowing the exception in `_send_and_forget` with a debug line is right — for a command the board was never going to answer, failing to get an answer is the expected ending, and there is nobody waiting to be told.

Documenting `timeout` as ignored rather than quietly accepting it is the honest choice. It would be worse to take the parameter and appear to honour it.

CI does not run for PRs targeting a feature branch, so the local run is the check: **787 tests, ruff, `ruff format --check` and mypy clean.**

Thanks for rebasing these yourself after #569 — I said I would do it and then thought better of reaching into your branch, given how the last attempt went. Labelled `bug`. Merging.

PR #571, opened 2026-08-03
Report HTTP failures as the exceptions the contract names

trailro:
Targets `http-transport`, same as #546, #558, #569 and #570.

**Three of three**, from reviewing this PR's code. Last in the order I would merge them; it touches `_send_prepared_command`, as #569 does, so one of the two will need a rebase — happy to do it.

## Two outcomes that arrive as the wrong thing

**A 200 whose body is not JSON escapes as `JSONDecodeError`.** It is a `ValueError`, so the `except (ClientError, TimeoutError)` a few lines below does not catch it, and it reaches a caller with no reason to expect it. This is the ordinary wrong-address mistake rather than an exotic one: a mistyped IP landing on a router's admin page answers `200 text/html`. `connect()` promises `NotConnectedError` when nothing answers; a config flow catching that and `RPCError` gets a traceback instead.

The `isinstance(payload, dict)` guard just below shows the intent already — it normalises replies that parse but are not objects. This covers the ones that do not parse.

**An HTTP 401 becomes a bare `RPCError`.** It is constructed directly rather than through `transport._rpc_error`, which is what maps code 401 to `Unauthorized`. A caller catching `Unauthorized` to re-prompt for a password — which ha-pitboss does — sees a generic failure. A Mongoose build with `rpc.auth_file` set refuses at the HTTP layer rather than in a JSON-RPC error frame.

## Left alone deliberately

The per-call `timeout=` is also not honoured: `ClientTimeout` is fixed at construction, so a longer per-call value is capped and `None` does not mean forever, and aiohttp's expiry is swallowed into `NotConnectedError` rather than the documented `TimeoutError`.

I did not fix it here because a correct fix has to decide how the constructor's default composes with `Transport.send_command`'s own `asyncio.timeout`, and `_send_prepared_command` cannot see the per-call value at all — the base does not pass it down. That is an API shape question rather than a bug to patch quietly. Happy to write it whichever way you want it.

## Testing

Two tests, both **failing on this PR as it stands**: a router-style HTML answer surfaces as `RPCError`, and a 401 as `Unauthorized`.

784 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -51,7 +51,7 @@ from typing import Any
 from aiohttp import ClientError, ClientSession, ClientTimeout
 
 from .exceptions import NotConnectedError, RPCError
-from .transport import DEFAULT_TIMEOUT, Transport
+from .transport import DEFAULT_TIMEOUT, Transport, _rpc_error
 
 _LOGGER = logging.getLogger("pytboss")
 
@@ -211,11 +211,28 @@ class HttpConnection(Transport):
                 self._url, json=cmd, timeout=self._timeout
             ) as resp:
                 if resp.status != 200:
-                    raise RPCError(
-                        f"{self._url} answered HTTP {resp.status}", resp.status
+                    # Built through `_rpc_error` so a 401 arrives as
+                    # `Unauthorized`, which is what a caller re-prompting for
+                    # a password catches and what a Mongoose build with
+                    # `rpc.auth_file` set answers with.
+                    raise _rpc_error(
+                        {
+                            "code": resp.status,
+                            "message": f"{self._url} answered HTTP {resp.status}",
+                        }
                     )
                 # Mongoose does not always set a JSON content type.
-                payload: Any = await resp.json(content_type=None)
+                try:
+                    payload: Any = await resp.json(content_type=None)
+                except ValueError as ex:
+                    # A 200 carrying something that is not JSON is not this
+                    # protocol at all -- a mistyped address landing on a
+                    # router's admin page answers exactly like that. Left
+                    # alone it escapes as `JSONDecodeError`, which is neither
+                    # documented here nor caught by callers.
+                    raise RPCError(
+                        f"{self._url} answered with something that is not JSON"
+                    ) from ex
         except (ClientError, TimeoutError) as ex:
             # The grill is unreachable rather than refusing the call. Reported
             # as a lost connection so callers treat it the way they treat a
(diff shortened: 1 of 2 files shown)


dknowles2:
Both outcomes confirmed, before and after:

```
before:  HTML 200: JSONDecodeError    HTTP 401: RPCError
after:   HTML 200: RPCError          HTTP 401: Unauthorized
```

Both mine, and the second is the more embarrassing: I built the error directly rather than through `_rpc_error`, which exists precisely to map 401 onto `Unauthorized` — three lines above the code I was writing. ha-pitboss catches `Unauthorized` to start a reauth flow, so on this transport a grill with `rpc.auth_file` set would have given the user a generic failure and no route to fixing it. #350 and #363 both turned on getting that exception to the right place; I broke it in a new transport while those were being reviewed.

The first is the ordinary mistyped-address case rather than an exotic one, and you are right that the `isinstance(payload, dict)` guard just below shows I had already thought about replies that are not what we expect — I covered the ones that parse and not the ones that do not.

**Leaving the per-call `timeout=` alone was the right call**, and the reason you give is the one that matters: `_send_prepared_command` cannot see the per-call value at all, because the base does not pass it down. So a correct fix changes `Transport`'s contract for every transport, and doing that quietly inside an HTTP bug-fix would be the wrong place to decide it. Worth its own issue once #543 lands — I would rather it were argued in the open than patched around.

787 tests, ruff and mypy clean on this branch.

**It went DIRTY again when #570 landed** — `tests/test_http.py` only, the same both-added collision as last time. Rebase when convenient and I will merge; the review above stands and I do not expect anything further. I am deliberately not reaching into your branch to do it, for the reason I gave on #570.

Labelled `bug`.

dknowles2 (approved):
Rebased cleanly and the `http.py` change is byte-identical to what I reviewed. **789 tests**, ruff, `ruff format --check` and mypy clean.

With all three landed, the four defects you found in this transport are gone together:

```
after disconnect(): NotConnectedError   (was: command SUCCEEDED, is_connected=True)
reboot:             returned in 0.00s   (was: NotConnectedError after the full timeout)
HTML 200:           RPCError            (was: JSONDecodeError)
HTTP 401:           Unauthorized        (was: bare RPCError)
```

Worth saying plainly: **you found four real bugs in code I wrote and reviewed myself**, three of them on paths I had convinced myself were covered. The session guard passed every test I wrote because for an owned session the two conditions coincide; the inherited fire-and-forget was correct for both existing transports; and the 401 mapping I broke sits three lines from the helper that exists to do it. That is not a run of bad luck, it is what reviewing your own transport gets you.

Merging. That is #546, #558, #569, #570 and #571 all landed against `http-transport` — #543 is now materially your work as much as mine, and I will say so when David looks at it.

PR #572, opened 2026-08-03
Serialize the websocket lifecycle, and cancel only the handshake

### LINKED ISSUE #567: WebSocketConnection has no lifecycle lock: connect() leaks a socket, disconnect() can hang
`BleConnection` grew a `_lifecycle_lock` for exactly this class of problem. `WebSocketConnection` has none, and two things follow. Raising as an issue rather than a PR because the fix is a design choice, and `ble.py` offers more than one precedent.

## Concurrent `connect()` leaves a live cloud socket nobody owns

Both calls pass the same checks (the subscribe task is still `None`), both open a socket, and both overwrite `_sock` and `_subscribe_task`. `_subscribe()` re-reads `self._sock` at its `async with`, so **both tasks end up driving the second socket** while the first is never entered, never iterated and never closed.

```
sockets opened by the server : 2 (expect 1)
live _subscribe tasks        : 2 (expect 1)
sockets still open after disconnect(): 1 (expect 0)
```

That last line is the part that matters: `disconnect()` cannot reach the orphan, so the session to `socket.dansonscorp.com` stays open for the life of the process.

Reachable from Home Assistant, which retries a config entry setup while a reload is in flight, and from any caller that calls `PitBoss.start()` twice.

## `disconnect()` hangs if the receive loop is inside `_ws_connect()`

`_stop_subscribing()` sets `_keep_running = False`, sets `_stopping`, closes `_sock` **only if it is not None**, then awaits the subscribe task. During a reconnect attempt `_sock` *is* `None` and the task is blocked in `session.ws_connect()`, which neither flag interrupts.

```
disconnect() DID NOT RETURN within 1.5s while a handshake was in flight
```

And when the handshake finally finishes, `_subscribe()` only tests `_keep_running` at the top of its `while`: it drops straight into `async with self._sock: async for msg in self._sock:` and starts serving a connection it was told to abandon.

Both probes are against `main` (a210df1), driven by a real aiohttp test server for the first and a stalled `_ws_connect` for the second.

## What I would want to know before writing it

The lock is the easy half. The harder half is interrupting a handshake in flight — cancelling the subscribe task rather than awaiting it changes what `disconnect()` guarantees to callers mid-command, and `_fail_pending_commands()` already covers part of that ground. `ble.py` cancels; `wss.py` awaits. Happy to send a PR once you say which way you want it, or to leave it with you.

Related but separate, and reasoned rather than reproduced: calling `PitBoss.stop()` from inside a state callback would make the receive-loop task await itself.

trailro:
Fixes #567, built to the design you set out there rather than the one I opened with.

## What changed from my original suggestion

You were right that cancelling the subscribe task would be a regression, and I checked it before writing this. On `main` today, a command in flight when `disconnect()` is called fails **immediately**:

```
in-flight command failed promptly after disconnect(): 0.00s   <- _fail_pending_commands ran
```

So `disconnect()` really does promise callers that the loop drained. Cancelling the task skips the line that does it, and every in-flight command would instead wait out its full timeout — a quieter failure than the hang it was meant to fix. `test_disconnect_still_fails_commands_in_flight` now pins that, so the trade cannot be made by accident later.

## The three pieces

**A lifecycle lock** on `connect`/`disconnect`, in the shape `ble.py` uses: public method takes it, `_locked` variant does the work.

**The handshake as its own task**, cancelled by `_stop_subscribing()` while the subscribe task is still awaited. `CancelledError` is re-raised when `_keep_running` is still set, so a genuine outside cancellation is not swallowed as a disconnect.

**The flag re-checked after the socket is assigned**, for the handshake that completes after the abandon order and beats the cancel.

Plus the `asyncio.current_task()` guard you suggested. asyncio raises there on its own, but only once the await is attempted, and with a message about a task awaiting itself — which is a long way from "a subscriber stopped the transport that was dispatching it".

## Verified

Both original probes, re-run against this branch:

```
A sockets still open after disconnect: 0        (was 1 — the orphan)
B disconnect() returned in 0.00s                (was: never returned)
B still connected afterwards? False
C in-flight command still failed promptly: 0.00s  (the drain, not regressed)
```

Five tests. **Four fail on `main`**; the fifth is the drain guard, which passes there because it is describing behaviour worth keeping.

One of them is written against `_subscribe()` directly rather than raced, because the cancel normally wins: it enters the loop and has the handshake flip the flag as it returns, which is exactly the state a losing cancel leaves behind.

799 pass, `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -83,6 +83,7 @@ class WebSocketConnection(Transport):
         :raise pytboss.exceptions.NotConnectedError: If a session was supplied
             by the caller and has since been closed.
         """
+        self._check_not_reentrant()
         async with self._lifecycle_lock:
             await self._connect_locked()
 
@@ -98,17 +99,25 @@ class WebSocketConnection(Transport):
         self._subscribe_task = self._loop.create_task(self._subscribe())
         await self._subscribed.wait()
 
+    def _check_not_reentrant(self) -> None:
+        """Refuse a lifecycle call made from a callback we are dispatching.
+
+        Awaiting the subscribe task from inside itself never returns, so say
+        so rather than hang. Checked *before* the lifecycle lock is taken:
+        past it, the caller would block on the lock while whoever holds the
+        lock awaits the very subscribe task this caller is running on -- a
+        deadlock the guard exists to prevent, not cause.
+        """
+        if self._subscribe_task is asyncio.current_task():
+            raise RuntimeError(
+                "Cannot stop this transport from a callback it is dispatching"
+            )
+
     async def _stop_subscribing(self) -> None:
         """Wind down a running subscribe task, if there is one."""
         if self._subscribe_task is None:
             return
-        if self._subscribe_task is asyncio.current_task():
-            # A subscriber stopping the transport that is dispatching it.
-            # Awaiting the task from inside itself never returns, so say so
-            # rather than hang.
-            raise RuntimeError(
-                "Cannot stop this transport from a callback it is dispatching"
-            )
+        self._check_not_reentrant()
         self._keep_running = False
         self._stopping.set()
         # The handshake is the one await neither flag reaches -- aiohttp
@@ -134,6 +143,7 @@ class WebSocketConnection(Transport):
         internally (i.e. no `session` was passed to `__init__`), and waits
         for the background reconnect/subscribe task to finish.
         """
+        self._check_not_reentrant()
         async with self._lifecycle_lock:
             await self._disconnect_locked()
 
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -61,6 +61,13 @@ class WebSocketConnection(Transport):
         self._subscribed = Event()
         self._stopping = Event()
         self._keep_running = False
+        self._connect_task: Task | None = None
+        # Serializes connect/disconnect. Opening a socket is an await, so two
+        # callers otherwise both pass the wind-down check, both open one, and
+        # the second assignment orphans the first -- along with its task,
+        # which `disconnect()` can no longer reach. Distinct from
+        # `_sock_lock`, which serializes sends.
+        self._lifecycle_lock = Lock()
 
     async def connect(self) -> None:
         """Starts the connection to the device.
@@ -72,6 +79,10 @@ class WebSocketConnection(Transport):
         :raise pytboss.exceptions.NotConnectedError: If a session was supplied
             by the caller and has since been closed.
         """
+        async with self._lifecycle_lock:
+            await self._connect_locked()
+
+    async def _connect_locked(self) -> None:
         await self._stop_subscribing()
         if self._session is None or self._session.closed:
             if not self._session_owned:
@@ -87,8 +98,22 @@ class WebSocketConnection(Transport):
         """Wind down a running subscribe task, if there is one."""
         if self._subscribe_task is None:
             return
+        if self._subscribe_task is asyncio.current_task():
+            # A subscriber stopping the transport that is dispatching it.
+            # Awaiting the task from inside itself never returns, so say so
+            # rather than hang.
+            raise RuntimeError(
+                "Cannot stop this transport from a callback it is dispatching"
+            )
         self._keep_running = False
         self._stopping.set()
+        # The handshake is the one await neither flag reaches -- aiohttp
+        # offers no way to interrupt `ws_connect` -- so it is cancelled
+        # directly. The subscribe task itself is still awaited rather than
+        # cancelled, so the `_fail_pending_commands()` below its read loop
+        # still runs and in-flight commands fail now instead of timing out.
+        if self._connect_task is not None:
+            self._connect_task.cancel()
         if self._sock is not None and not self._sock.closed:
             await self._sock.close()
         await self._subscribe_task
@@ -104,6 +129,10 @@ class WebSocketConnection(Transport):
         internally (i.e. no `session` was passed to `__init__`), and waits
         for the background reconnect/subscribe task to finish.
         """
+        async with self._lifecycle_lock:
+            await self._disconnect_locked()
+
+    async def _disconnect_locked(self) -> None:
         # Wakes the subscribe task if it is waiting out a reconnect backoff;
         # otherwise this call blocks for the remainder of that sleep.
         await self._stop_subscribing()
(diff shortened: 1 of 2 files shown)


dknowles2 (commented):
The design is right and the two original defects are fixed. I confirmed your headline claim the way you stated it — the five new tests against `main`'s `wss.py`:

```
4 failed, 1 passed
```

and the one that passes there is `test_disconnect_still_fails_commands_in_flight`, exactly as you said: it describes behaviour worth keeping rather than behaviour being added. 799 pass on the branch, `ruff check`, `ruff format --check` and `mypy .` clean.

Checking that you were right about the drain before writing the code, rather than taking my word for it, is the right instinct — and it is what makes the `break`-vs-`cancel` decision defensible instead of just agreed.

## The handshake cancel is not pinned, and finding out why turned up something

I mutated each mechanism separately:

```
lock off connect()            -> test_concurrent_connects_leave_one_live_socket   FAILS
lock off disconnect()         -> 27 passed
no handshake cancel           -> 27 passed
no post-handshake re-check    -> test_a_handshake_that_beats_the_wind_down...     FAILS
no current_task() guard       -> test_stopping_from_a_dispatched_callback...      FAILS
swallow an outside cancel     -> 27 passed
```

`test_disconnect_does_not_wait_out_a_handshake` passes with the cancel removed. That surprised me, because it fails on `main`, so I probed it rather than reporting it:

```
PR as written     disconnect() returned normally after 0.00s
cancel removed    disconnect() returned normally after 2.00s
```

The fix works — 0.00s is the whole point of the PR. But with the cancel gone, `disconnect()` **waits out the caller's entire `asyncio.timeout(2)` and then returns normally instead of raising `TimeoutError`.** The test's only bound is that timeout, and the timeout is being absorbed, so the test cannot tell the two apart.

## Why it is absorbed, which is the part worth fixing

```python
except asyncio.CancelledError:
    if self._keep_running:
        raise
    break
```

The guard distinguishes "cancelled from outside" from "cancelled by the wind-down" using `_keep_running` — but `_stop_subscribing()` sets `_keep_running = False` *before* it awaits the subscribe task. So during a disconnect the flag reads False whoever did the cancelling, and an externally-delivered cancellation arriving in that window is swallowed and turned into a clean `break`.

This does not exist on `main`: there is no `except asyncio.CancelledError` there at all, so a caller's cancellation propagates. So the PR introduces it. It is the same shape as the bug you filed — a caller asks for a bound, does not get it, and is told everything is fine — and it is reachable by anyone who wraps `disconnect()` in a timeout, which is a reasonable thing to do to a method whose docstring says it waits for a background task.

Distinguishing on a flag the wind-down has already flipped cannot work; it needs to key on whether *this* code did the cancelling. A dedicated flag set immediately before `.cancel()` and cleared in the handler is the smallest version:

```python
# _stop_subscribing
if self._connect_task is not None:
    self._handshake_cancelled = True
    self._connect_task.cancel()

# _subscribe
except asyncio.CancelledError:
    if not self._handshake_cancelled:
        raise
    self._handshake_cancelled = False
    break
```

Whatever shape you pick, the test wants an elapsed-time assertion rather than only the `timeout()`, or it will keep passing on a version that hangs.

## Smaller

`disconnect()`'s lock acquisition is unpinned, same as it was on #553 — I said then it was worth a follow-up and it still is, not a blocker.

`test_a_handshake_that_beats_the_wind_down_is_not_served` leaks an aiohttp session:

```
Unclosed client session
```

The `session` fixture returns a bare `ClientSession()` and never closes it; the `conn` fixture happens to close it with `async with`, and this is the first test to take `session` without `conn`. Pre-existing shortcoming in the fixture, newly visible. It prints on every run now, which is how a real leak gets missed later. Closing it in the fixture fixes it for good.

## Where that leaves it

Everything you set out to do works, and the `current_task()` guard message is a genuine improvement on asyncio's — "a task awaiting itself" tells a user nothing about what they did. My one blocking concern is the swallowed cancellation, because it is new behaviour rather than an uncovered path, and it is the kind of thing that is much cheaper to fix now than to find later in a release.

trailro:
You are right, and I reproduced it before changing anything rather than taking the mutation result on faith:

```
outside cancellation SWALLOWED: task finished normally
```

A wind-down in progress, a cancellation delivered by something else, and `_subscribe()` hands back a clean return. Your diagnosis of why is exact: `_stop_subscribing()` clears `_keep_running` before it cancels, so the flag cannot answer the question I was asking it. And it is new — `main` has no `CancelledError` handler at all — so I introduced it while fixing the hang. Same shape as the bug I filed, which is not lost on me.

Fixed with the dedicated flag, with one addition: it is cleared in the `finally` rather than in the handler. A cancel that never lands — the handshake finishing first — would otherwise leave it set and make the *next* cancellation look like ours.

```
outside cancellation propagated correctly
```

**The test now asserts elapsed time as well as the timeout**, as you said it needed to. Checked both ways: reverting the flag fails `test_a_cancellation_we_did_not_ask_for_is_not_swallowed`, and removing the cancel fails `test_disconnect_does_not_wait_out_a_handshake` on the clock rather than passing at 2.00s.

**The session fixture closes what it opens.** You are right that it is pre-existing and newly visible; I would rather not leave a warning printing on every run, since that is exactly how the next one gets missed. No "Unclosed client session" left in the file's output.

`disconnect()`'s unpinned lock acquisition I have left alone, as a follow-up rather than a blocker, same as on #553.

800 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

Two things worth saying plainly. Mutating each mechanism separately is what found this — my five tests all passed, and four of them failed on `main`, which is exactly the evidence that makes a change look done. It was not. And the failure mode you describe is the one I would least like to ship: not a crash, but a caller being told its bound was honoured when it was not.

trailro:
Rebased onto `main` (the branch predated #543/#458/#571 landing), plus one more fix from continued adversarial testing of my own change.

The reentrancy guard was checked inside `_stop_subscribing()` — after the lifecycle lock was taken. Whenever another lifecycle call was in flight, a subscriber calling `disconnect()` blocked on the lock and never reached the guard, while the lock's holder awaited the very subscribe task dispatching that subscriber. Sampled mid-hang at t=2s with nothing cancelled: lock held, one waiter, all three tasks incomplete, permanently. `main` self-heals from the same sequence (the self-await `RuntimeError` is swallowed by the read loop and things unwind) — so the guarded version was strictly worse than the bug it replaced in this scenario.

The guard now runs at the top of `connect()`/`disconnect()`, before the lock, and stays in `_stop_subscribing()` as a backstop for direct callers. The new test hangs (fails on its 3s bound) with the guard in the old place and passes with it moved; verified both ways.

875 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

dknowles2 (approved):
The blocking item from the earlier review is fixed: `_handshake_cancelled` now disambiguates an externally-delivered cancellation from the wind-down's own, and the swallow only fires on ours. Verified with a worktree checkout — `test_a_cancellation_we_did_not_ask_for_is_not_swallowed` and `test_disconnect_does_not_wait_out_a_handshake` both exercise the race directly, and all 29 tests in `test_wss.py` pass.

Bonus: `_lifecycle_lock` also serializes connect/disconnect, which independently addresses the unpinned-lock note from #553's follow-up — nice to get both in one pass. Session fixture leak closed too.

PR #573, opened 2026-08-04
Wrap the remaining RPCs the firmware and Mongoose document

dknowles2:
Closes the remaining wrappable items on #167. Four methods, each written against a source rather than against the list in that issue.

| Method | RPC | Grounded in |
|---|---|---|
| `rename_device()` | `PB.RenameDevice` | `fake_firmware/init.js:668` |
| `set_wifi_credentials()` | `PB.SetWifiCredentials` | `fake_firmware/init.js:637` |
| `debug_pstate()` | `PB.DebugPState` | `fake_firmware/init.js:614` |
| `scan_wifi()` | `Wifi.Scan` | [Mongoose `rpc-service-wifi`](https://mongoose-os.com/docs/mongoose-os/api/rpc/rpc-service-wifi.md) |

## What the firmware actually does, which is not what the names suggest

**`PB.RenameDevice` does not set the name you give it.** It keeps everything up to and including the first `-` in the device id and appends the argument, so `PBL-1234` renamed to `Patio` becomes `PBL-Patio` — and the reply is that whole id, not the name. `rename_device()` returns it for that reason.

It also **rejects rather than trims**: an empty name, or one with a leading or trailing space, falls through to `Invalid parameters`. Both are pinned.

**`PB.SetWifiCredentials` uses a second codec key**, `[0x8f, 0x80, 0x27, 0xcf, 0x41, 0x6c, 0x45, 0xb7]` — four bytes different from the grill-password key and not time-varied. It is now `WIFI_KEY` in `codec.py`, next to `KEY`, with the difference documented so neither gets used for the other.

That is the only thing it offers over `Config.set_wifi_credentials()`, which reaches the same `wifi.sta` settings through Mongoose's config service and sends the password as plain text. Both are kept and the docstrings point at each other.

**`PB.DebugPState` is written only by the cloud.** A `setPState` frame on the outbound socket sets it; the grill never does, and echoes it back on every status push. So it reports what the vendor's service last said, and is empty on a grill reached over Bluetooth or the local transport.

## On `Wifi.Scan`

Not in the application — it is a Mongoose core service — so there is no handler to read and the shape is the documentation's: a bare array of `{ssid, bssid, auth, channel, rssi}`, no parameters.

Worth recording that I had this wrong first: I took the shape from the [JS `Wifi.scan()` page](https://mongoose-os.com/docs/mongoose-os/api/net/wifi.md), which names the field `authMode`. The RPC service names it `auth`. Same networks, different API. @dknowles2 pointed at the right page.

Returned as sent rather than remapped, and guarded against a non-list exactly as `list_rpcs()` is, since nothing here has seen a grill answer it.

## Not wrapped, deliberately

**`PBL.StartWifiScan` and `PBL.GetWifiScanStatus`.** No `PBL.*` handler appears in the application at all, and neither method is documented anywhere I can reach. There is nothing to write them against, and writing them from the name is how `OTA.Start` got into #458.

The rest of #167's list stays declined for the reasons @trailro gave on the issue: `PB.LoadMCUFirmware` and friends are ESP-IDF-only, and the `PBL.LoadFirmware` trio wraps a path the vendor disabled in 0.5.7 while `CopyFile` remains a live way to overwrite files on a running grill.

## Verified

**842 tests** (up from 819), `ruff check`, `ruff format --check` and `mypy .` clean.

Mutated each mechanism separately; every one is pinned:

```
rename: drop the password        -> test_rename_device_keeps_the_prefix[with_password]  FAILS
rename: return name not id       -> test_rename_device_keeps_the_prefix                 FAILS
wifi: use the grill-password key -> test_set_wifi_credentials_obfuscates_the_password   FAILS
wifi: send the password as-is    -> test_set_wifi_credentials_obfuscates_the_password   FAILS
pstate: drop the password        -> test_debug_pstate[with_password]                    FAILS
scan: drop the list guard        -> test_scan_wifi_when_the_grill_does_not_serve_it     FAILS
```

**Not exercised on hardware.** The fake transport mirrors the firmware's handlers, including the rejection paths, but no grill has answered any of these four. `rename_device()` changes the device id, which is what identifies a grill elsewhere, so that is the one worth trying deliberately rather than incidentally.

### DIFF
--- a/pytboss/__init__.py
+++ b/pytboss/__init__.py
@@ -4,4 +4,5 @@ from .api import PitBoss  # noqa: F401
 from .ble import BleConnection  # noqa: F401
 from .http import HttpConnection  # noqa: F401
 from .ota import OTA  # noqa: F401
+from .wifi import WiFi  # noqa: F401
 from .wss import WebSocketConnection  # noqa: F401
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -675,6 +675,68 @@ class PitBoss:
         result = await self._conn.send_command("Wifi.Scan", {})
         return result if isinstance(result, list) else []
 
+    async def start_wifi_scan(self) -> dict:
+        """Asks the loader to begin scanning for WiFi networks.
+
+        Answers `{"scanning": bool, "results": list | None}` immediately
+        rather than waiting. A scan already in flight is left alone and its
+        state returned, so calling this twice does not restart anything.
+
+        The loader serves this, not the grill application, and does not check
+        the password. Prefer `scan_wifi_networks()` unless you want to drive
+        the polling yourself -- `get_wifi_scan_status()` hands the results
+        back only once.
+        """
+        return as_dict(await self._conn.send_command("PBL.StartWifiScan", {}))
+
+    async def get_wifi_scan_status(self) -> dict:
+        """Returns the state of the scan `start_wifi_scan()` began.
+
+        **Reading the results consumes them.** The loader clears its stored
+        results as it returns them, so the next call answers `None` again
+        even though the scan completed. Whatever calls this owns the only
+        copy.
+
+        `{"scanning": True, "results": None}` while it runs, then
+        `{"scanning": False, "results": [...]}` exactly once.
+
+        Each entry carries `ssid`, `bssid`, `authMode`, `channel` and `rssi`
+        -- `authMode`, because the loader goes through Mongoose's JS
+        `Wifi.scan()`, where `scan_wifi()` calls the `Wifi.Scan` RPC and gets
+        the same field named `auth`. Same grill, same networks, two spellings.
+        """
+        return as_dict(await self._conn.send_command("PBL.GetWifiScanStatus", {}))
+
+    async def scan_wifi_networks(
+        self, *, timeout: float = 15.0, poll_interval: float = 1.0
+    ) -> list[dict]:
+        """Runs the loader's scan to completion and returns what it found.
+
+        The two-call dance in one place, so the single destructive read of
+        `get_wifi_scan_status()` happens once, here, rather than in every
+        caller that polls.
+
+        Empty if the scan finds nothing, if the grill serves no loader, or if
+        `timeout` elapses first -- none of which are distinguishable in the
+        reply, so this does not pretend to tell them apart.
+
+        :param timeout: Seconds to keep polling before giving up.
+        :param poll_interval: Seconds between polls.
+        """
+        await self.start_wifi_scan()
+        deadline = monotonic() + timeout
+        while monotonic() < deadline:
+            await asyncio.sleep(poll_interval)
+            status = await self.get_wifi_scan_status()
+            results = status.get("results")
+            if isinstance(results, list):
+                return results
+            if not status.get("scanning"):
+                # Not running and nothing held: either it never started or a
+                # previous caller already consumed the results.
+                return []
+        return []
+
     async def rename_device(self, name: str) -> str:
         """Renames the grill, returning its new device id.
 
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -16,6 +16,7 @@ from .fs import FileSystem
 from .grills import Grill, StateDict, get_grill
 from .ota import OTA
 from .transport import RPCResult, Transport, as_dict
+from .wifi import WiFi
 
 _UPTIME_TTL = 60.0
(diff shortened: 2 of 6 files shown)


dknowles2:
Added the loader's WiFi scan pair, which I had declined in the original description for want of a source. @dknowles2 supplied `app.js` — the **PitBoss Loader 0.2.2**, where every `PBL.*` handler lives — so they are now written against the handlers rather than against their names.

* `start_wifi_scan()` → `PBL.StartWifiScan`
* `get_wifi_scan_status()` → `PBL.GetWifiScanStatus`
* `scan_wifi_networks()` → both, driven to completion

## Three things the names do not tell you

**`PBL.GetWifiScanStatus` is a destructive read.** It copies its stored results, sets them back to `None`, and returns the copy — so a completed scan can be read exactly once, and whoever polls owns the only copy. A second poll answers `None` as though the scan never finished. `scan_wifi_networks()` exists so that single read happens in one place instead of in every caller.

**The loader calls Mongoose's JS `Wifi.scan()`, not the `Wifi.Scan` RPC.** So its entries carry `authMode`, where `scan_wifi()` — added in the first commit — gets the same field named `auth`. Same grill, same networks, two spellings, and there is now a test asserting both rather than a comment claiming it.

**None of the six `PBL.*` handlers checks the password.** Worth stating plainly since the other RPCs in this PR do.

## Still not wrapped, and now for a documented reason rather than an absent one

`PBL.LoadFirmware` and `PBL.LoadFirmwareStatus`: `downloadInfo()` opens with

```js
fw_status = "error-downloading-file";
fw_error = true;
return;
```

and the rest of the function is unreachable. So on this loader the load path always fails, which confirms what @trailro reported on #167 from the 0.5.7 images.

`PBL.CopyFile` is the opposite — live, unauthenticated, and it copies any file over any other on a running grill. Wrapping that is a decision rather than an omission, so I have left it.

## Verified

**857 tests**, `ruff check`, `ruff format --check`, `mypy .` clean.

Mutated each mechanism separately:

```
drop the early stop when not scanning  -> test_scan_wifi_networks_stops_when_the_grill...  FAILS
never call start                       -> test_scan_wifi_networks_drives_the_whole_dance   FAILS
ignore the results when they arrive    -> test_scan_wifi_networks_drives_the_whole_dance   FAILS
drop the deadline                      -> HUNG, which is the failure it should be
```

One correction to my own test while writing these: I first wrote a case asserting `scan_wifi_networks()` returns empty after another caller consumed the results. That was wrong — it starts its own scan, so a consumed result is irrelevant. The test now pins that behaviour instead, and the early-stop branch is covered separately with a transport that reports idle.

**Not exercised on hardware.** The fake mirrors the handlers, including the destructive read, but no grill has answered any of this.

dknowles2:
Moved `Wifi.Scan` onto its own class, per @dknowles2 — it is a Mongoose core service with nothing PitBoss about it, and this library already splits by RPC service rather than by topic (`fs.py` is `FS.*`, `ota.py` is `OTA.*`, `config.py` is `Config.*`).

```
boss.scan_wifi()  ->  boss.wifi.scan()
```

Never released, so there is nothing to keep working.

## The `PBL.*` pair stays on `PitBoss`, deliberately

It reads like the same topic and is not the same service. `PBL` is the **PitBoss Loader** — a separate Dansons application with its own version, not Mongoose. And grouping by topic rather than service would also mean moving `Config.set_wifi_credentials()`, `set_wifi_ssid()` and `set_wifi_password()`, which live on `Config` because they go through `Config.Set`.

So this PR now has WiFi methods in two places, which is the honest outcome of splitting by service: `boss.wifi.scan()` for Mongoose's, `boss.scan_wifi_networks()` for the loader's. Each docstring points at the other and names the field-spelling difference (`auth` vs `authMode`) that makes them distinguishable in practice.

A `Loader` class collecting `PBL.GetLoaderVersion`, the scan pair, and eventually `CopyFile` is the coherent next step. It would move `get_loader_version()`, which **2026.8.4 already shipped**, so that is a decision about breaking released API rather than a cleanup — worth its own PR.

## Verified

**858 tests**, `ruff check`, `ruff format --check`, `mypy .` clean.

Checked the move rather than assuming it:

```
boss.wifi is a WiFi:              True
scan_wifi gone from PitBoss:      True
loader pair still on PitBoss:     True
```

`tests/test_wifi.py` is new and covers the non-list guard three ways — `None`, an object, and a genuinely empty result — since "found nothing" and "does not serve it" are indistinguishable in the reply and only one of them is a bug if it regresses.

PR #575, opened 2026-08-04
Survive the network failures a reconnect actually meets

trailro:
## What

`WebSocketConnection`'s reconnect loop only survives `GrillUnavailable`, and `_ws_connect` raises that only for `WSServerHandshakeError`. The failures a reconnect actually meets are different ones: an unreachable relay raises `ClientConnectorError`, a dropped network `OSError`, a stalled TCP handshake `TimeoutError`. Each of those escapes the loop's `except GrillUnavailable` and kills the subscribe task — permanently ending the automatic reconnection the class docstring promises, with nothing surfaced to the consumer.

The dead task then poisons `disconnect()`: `_stop_subscribing` awaits it bare, so the task's exception re-raises on **every** later `disconnect()` call, forever. The lines below the await never run, so an internally-created `ClientSession` is never closed. Observed:

```
disconnect() #1: ClientConnectorError: Cannot connect to host ...
disconnect() #2: ClientConnectorError: Cannot connect to host ...
owned session closed: False
```

So a plain Wi-Fi blip leaves the transport dead, un-disconnectable, and leaking.

## Fix

Two halves, same root:

- `_ws_connect` maps `ClientError`, `OSError` and `TimeoutError` to `GrillUnavailable` — everything the network can do there means the same thing: the grill cannot be reached right now. (`WSServerHandshakeError` is a `ClientError`, so the existing case is subsumed.)
- `_stop_subscribing` retrieves a dead task's error defensively (logged, not re-raised) and clears its state in a `finally`, so `disconnect()` always completes and the owned session always closes.

## Tests

Both fail against unpatched `main`, verified:

- `test_reconnect_survives_the_grill_vanishing_from_the_network` — after the stream ends and reconnects target a dead port, the loop must still be retrying; on main it dies with `ClientConnectorError`.
- `test_a_dead_subscribe_task_does_not_poison_disconnect` — a task killed by an unanticipated error must not stop `disconnect()` from completing and closing the owned session; on main the failure even prints aiohttp's "Unclosed client session" warning.

870 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -8,10 +8,10 @@ from typing import Any
 from uuid import uuid4
 
 from aiohttp import (
+    ClientError,
     ClientSession,
     ClientWebSocketResponse,
     WSMsgType,
-    WSServerHandshakeError,
 )
 
 from .exceptions import GrillUnavailable, NotConnectedError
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Verified against fake_firmware and the diff: WSServerHandshakeError only maps the handshake-error case, and ClientConnectorError/OSError/TimeoutError are exactly what an unreachable relay or dropped network raise in practice, escaping the old `except GrillUnavailable`. The dead-task fix in `_stop_subscribing` (log-and-continue instead of re-raising forever) is also correct and covered. Tests target both scenarios directly.

dknowles2:
Approved above. Merging #576/#577/#580 first (independent fixes, all touched other files) left this one conflicting with #572's already-merged additions at the tail of `tests/test_wss.py` — both append new tests at the end of the file. Not a defect in this PR, just merge ordering on my end. A rebase onto `main` should resolve it cleanly; happy to do that myself if you'd rather not, or wait for you to push one.

trailro:
Rebased onto `main` — the tail-of-file test conflict resolved by keeping both blocks, and the merged result is coherent: #572's reentrancy guard and this PR's defensive dead-task retrieval coexist in `_stop_subscribing` as intended. 883 pass, ruff/format/mypy clean, CI green.

PR #576, opened 2026-08-04
Keep the HTTP transport honest when its own deadline fires first

trailro:
## What

Two holes in `HttpConnection`'s state-keeping, both only visible at the configuration every real consumer runs — the defaults.

**1. `is_connected()` lies forever once the grill goes silent.** `Transport.send_command` arms `asyncio.timeout(30.0)` and the transport builds `ClientTimeout(total=30.0)` — but aiohttp ceil-rounds its deadline up to the next whole second for timeouts ≥ 5s, so the RPC deadline always fires first. The request is *cancelled* rather than timing out inside the `except (ClientError, TimeoutError)` that clears `_connected`, so that branch is dead code at the defaults. Measured against a grill that answered `RPC.Ping` then went silent:

```
after connect(): is_connected()=True
  raised TimeoutError() after 30.0s
after the grill went silent: is_connected()=True   <- forever
```

A coordinator polling `is_connected()` keeps reporting the grill online after it stopped answering.

**2. A cancelled `connect()` skips the rollback and leaks the session.** `except Exception` does not catch `CancelledError`, so a caller-imposed cancellation — a config flow timing out around `connect()`, exactly what Home Assistant does — leaves `_started=True`, `_connected=True`, `is_connected()=True`, and the just-opened owned `ClientSession` unclosed (aiohttp prints `Unclosed client session`), once per retry.

Related: at those same default deadlines, `connect()` against a silent grill raised a bare `TimeoutError` where its docstring promises `NotConnectedError`.

## Fix

- `_send_prepared_command` gains a `CancelledError` branch that clears `_connected` before re-raising — the cancellation that reaches it is the RPC deadline firing mid-request, which means precisely "the grill did not answer".
- `connect()`'s rollback catches `BaseException`, so cancellation rolls back too, and maps `TimeoutError` to the `NotConnectedError` it documents.

## Tests

All three fail against unpatched `main`, verified:

- `test_is_connected_stops_lying_when_the_grill_goes_silent` — answered once, then quiet; after the RPC deadline fires, `is_connected()` must be `False`.
- `test_a_cancelled_connect_rolls_back` — cancelled mid-ping: not connected, owned session closed.
- `test_connect_maps_a_bare_timeout_to_the_documented_error` — a `TimeoutError` out of the RPC layer surfaces as `NotConnectedError`.

871 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -105,10 +105,24 @@ class HttpConnection(Transport):
         self._connected = True
         try:
             await self.send_command("RPC.Ping", {})
-        except Exception:
+        except BaseException as ex:
+            # `BaseException`, not `Exception`: a caller-imposed cancellation
+            # -- a config flow timing out around `connect()` -- otherwise
+            # skips this rollback entirely, leaving the transport claiming to
+            # be connected and the owned session it just opened leaking, once
+            # per retry.
             self._started = False
             self._connected = False
             await self._close_session()
+            if isinstance(ex, TimeoutError):
+                # At `send_command`'s own default deadline the RPC timeout
+                # fires before aiohttp's, so the unreachable-grill mapping in
+                # `_send_prepared_command` never runs. Map it here too, or
+                # this method answers a silent grill with a bare
+                # `TimeoutError` instead of the error it documents.
+                raise NotConnectedError(
+                    f"Could not reach {self._url}: no answer"
+                ) from ex
             raise
 
     async def disconnect(self) -> None:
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Checked the two failure windows against `Transport.send_command`'s timeout and aiohttp's ceil-rounding — the ordering claim (30.0s RPC deadline firing before aiohttp's own) holds, so both the `is_connected()` staleness and the swallowed-cancellation-on-connect were real. Catching BaseException here (not Exception) is required and correctly still re-raises after the rollback.

PR #577, opened 2026-08-04
Strip the password from pushed virtual data too

trailro:
## What

`get_virtual_data()` strips `psw` — the firmware stores the params of the last authenticated write verbatim, so the encoded password sits in the vdata scratchpad. But the **push** path doesn't: `_authenticate` injects `psw` into the very params object `PB.SetVirtualData` hands to the firmware (`vData = params` in `init.js`), and the firmware then re-broadcasts that object — `sendBTData` prints it whole to the BLE debug log, `sendWSStatus` attaches it as `data` on the status frame. Both land in `_on_vdata_received`, which passed it to every subscriber unfiltered:

```
subscriber received: {'p1T': 165, 'psw': 'a1b2c3deadbeef'}
```

Concrete path: any consumer that calls `set_probe_target()` on a board without a probe command takes the vdata route, and the grill echoes the encoded password back on the next push. A subscriber that logs or persists vdata records it — and the encoding is not protection, since the key is `timed_key(uptime)` and this library implements both ends.

## Fix

The same strip `get_virtual_data()` applies, applied in `_on_vdata_received` before dispatch (and before the DEBUG log line, which was also printing it). `get_virtual_data()`'s "this is the only thing that removes it" docstring updated to match.

## Test

`test_on_vdata_received_strips_the_password` covers both arrival shapes — the decoded object `wss` hands over and the JSON text `ble` reads off the debug log. Fails against unpatched `main`, verified.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -210,6 +210,16 @@ class PitBoss:
         # text the firmware printed; `wss` receives it as a member of an
         # already-parsed status frame, so it arrives decoded.
         vdata = json.loads(payload) if isinstance(payload, str) else payload
+        if isinstance(vdata, dict) and "psw" in vdata:
+            # The same strip `get_virtual_data()` applies, for the same
+            # reason: an authenticated write leaves the encoded password in
+            # the firmware's scratchpad, and the firmware pushes that object
+            # back verbatim -- on the BLE debug log and as `data` on the
+            # websocket status frame. Without this, every subscriber is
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Straightforward: the push path (`_on_vdata_received`) missed the same psw strip `get_virtual_data()` already applies. Test proves both dict and JSON-string payload shapes are stripped.

PR #578, opened 2026-08-04
Answer a board without the command with the documented error

trailro:
## What

`set_probe_temperature()` sends `set-probe-1-temperature` unguarded, while its sibling `set_probe_2_temperature()` checks membership and raises the documented `UnsupportedOperation`. Only 42 of 137 supported models declare the probe-1 slug, so on the other 95 the call crashes with a bare `KeyError('set-probe-1-temperature')` out of `_send_command`'s dict lookup — verified by sweeping every model in the catalogue:

```
set_probe_temperature    -> KeyError: 'set-probe-1-temperature'   (95 models, e.g. LG1200FL, all 32 PBC models)
set_probe_2_temperature  -> UnsupportedOperation                  (the sibling, same grill)
```

`turn_primer_motor_on()`/`turn_primer_motor_off()` have the same hole on the 26 models whose boards lack the primer slug (PBD, PBE, PBM, PBM2, PBV2, PBVA). For contrast, `turn_light_on()` is guarded via `has_lights` — and I cross-checked that no model with lights sits on a board lacking the slug, so that guard is sound as-is.

## Fix

All three methods now guard the way the probe-2 sibling does: membership check, `UnsupportedOperation` with a `:raise:` line in the docstring. `probe_target_command()` already documents the catalogue counts, so the docstrings carry them too.

## Test

`test_board_gated_commands_never_raise_a_bare_keyerror` sweeps every supported model in the catalogue and accepts only the documented outcomes. Fails against unpatched `main` (KeyError on the first affected model), verified.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -301,10 +301,15 @@ class PitBoss:
         """Sets the target temperature for probe 1.
 
         :param temp: Target probe temperature, in the grill's own unit.
+        :raise pytboss.exceptions.UnsupportedOperation: When probe 1's
+            target temperature cannot be set. Only 42 of 137 models declare
+            the command; on the rest this raised a bare `KeyError` while its
+            sibling below raised the documented error.
         """
-        return await self._send_command(
-            "set-probe-1-temperature", temp, self._is_fahrenheit()
-        )
+        cmd = "set-probe-1-temperature"
+        if cmd not in self.spec.control_board.commands:
+            raise UnsupportedOperation
+        return await self._send_command(cmd, temp, self._is_fahrenheit())
 
     async def set_probe_2_temperature(self, temp: int) -> RPCResult:
         """Sets the target temperature for probe 2.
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Matches the existing sibling pattern (`set_probe_2_temperature`) exactly. The test sweep executes across the full grill catalogue (checked > 100) rather than asserting from a read of grills.json, which is the right way to back this claim.

dknowles2:
Approved above. #577 merged first and also touched `pytboss/api.py`, so this now conflicts on rebase — again ordering on my end, not a problem with the change itself. Rebase onto `main` should be a clean resolve since the two touch different methods. Let me know if you'd like me to push that rebase, or I can wait for you.

trailro:
Rebased onto `main` — the conflict was only the tail of `tests/test_api.py` (#577's test and this PR's sweep both append there); both kept. 882 pass, ruff/format/mypy clean, CI green.

PR #579, opened 2026-08-04
Route a status push by its frame's prefix, not its position

trailro:
## What

`_handle_message` unpacks the status push positionally — `await self._state_callback(*payload["status"])` — assuming element 0 is always the FE0B state frame. The firmware builds that array conditionally (`init.js`):

```js
if (lastStatus.sc_11 !== "") wsStatus.push(toHexStr(lastStatus.sc_11));
if (lastStatus.sc_12 !== "") wsStatus.push(toHexStr(lastStatus.sc_12));
if (wsStatus.length < 1) return;
```

so `["FE0C…"]` is a producible frame. And the window is not exotic: `sendMCUCommand` blanks **both** `sc_11` and `sc_12` on every user-issued command, and `handlePacket` refills them one packet at a time — if the push timer fires after the FE0C reply lands but before the FE0B one, the grill sends `{"status": ["FE0C…"]}`. The same applies at boot before the first FE0B arrives.

Positionally unpacked, that frame arrives as the *status* payload; every board's `status_function` requires `startsWith("FE0B")` and returns nothing, so the temperatures are silently dropped and no subscriber is notified — precisely in the moments after the user changed something. Reproduced end-to-end:

```
frames=['FE0C..'] as positional arg 0: subscriber updates = 0
frames=['FE0C..'] routed to arg 1:     subscriber updates = 1, p1Temp=150
```

## Fix

A single-element push whose frame carries the FE0C prefix is routed to the temperatures argument — the same prefix dispatch `ble._on_debug_log_received` already does (`match payload[:4]: case "FE0B" / "FE0C"`). Everything else is unchanged, including the two-frame case and non-hex payloads.

## Test

`test_a_temperatures_only_push_reaches_the_callback_as_temperatures` covers both the temps-only push and the two-frame push through a real websocket round-trip. Fails against unpatched `main`, verified.

### DIFF
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -288,7 +288,24 @@ class WebSocketConnection(Transport):
                 await self._vdata_callback(vdata)
             if not self._state_callback:
                 return
-            await self._state_callback(*payload["status"])
+            frames = payload["status"]
+            if (
+                len(frames) == 1
+                and isinstance(frames[0], str)
+                and (frames[0][:4] == "FE0C")
+            ):
+                # The firmware builds this array conditionally -- a frame is
+                # pushed only if it is non-empty, and `sendMCUCommand` blanks
+                # both frames on every command, refilled one packet at a
+                # time. So a push carrying only the temperatures frame is
+                # producible, right in the window after a user-issued
+                # command, and positional unpacking would hand it over as
+                # the *status* payload -- where every board's status routine
+                # requires an FE0B prefix and returns nothing. Routed by the
+                # frame's own prefix instead, the way `ble` already does.
+                await self._state_callback(None, frames[0])
+                return
+            await self._state_callback(*frames)
             return
 
         if "id" in payload:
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Confirmed against fake_firmware/init.js lines 178-192 and 318-319: wsStatus is built conditionally and sc_11 is blanked between commands, so a temperatures-only push is genuinely producible, not a hypothetical. Routing by prefix instead of position is the same approach ble.py already uses. Good catch.

dknowles2:
Approved above. Same story as #575: merging other PRs first left this conflicting with #572's tests appended at the tail of `tests/test_wss.py`. Rebase onto `main` should resolve cleanly. Happy to push that myself (maintainer-edits is enabled on this branch) if you'd rather not — just say the word — or I can leave it for you.

trailro:
Rebased onto `main` — same tail-of-file resolve as #575, both test blocks kept. 882 pass, ruff/format/mypy clean, CI green.

dknowles2:
Approved and CI was green, but merging #575/#578 first (both landed cleanly) put this one in conflict — same tail-of-file pattern #575 hit against #572. A rebase onto `main` should resolve it; happy to do it myself or wait for you to push one.

trailro:
Rebased onto `main` — same tail-of-file resolve, #575's tests and this one's both kept. 885 pass, ruff/format/mypy clean.

PR #580, opened 2026-08-04
Bind the BLE client under the lock it is used with

trailro:
## What

Both GATT paths in `BleConnection` check `self._ble_client` outside the lock and dereference it inside — and `_on_disconnected` sets it to `None`, so every `await` between chunks is a window. Reproduced both:

- **Send** (`_send_prepared_command`): a drop after the length header is written but before the payload chunks raises `AttributeError: 'NoneType' object has no attribute 'write_gatt_char'` instead of the `NotConnectedError` the transport contract promises — and that this method's own comment insists on.
- **Receive** (`_on_rpc_data_received`): a drop between chunks of a multi-chunk reply raises `AttributeError: 'NoneType' object has no attribute 'read_gatt_char'` inside bleak's notification dispatch, where it surfaces as an unhandled-task traceback rather than reaching any caller.

This is precisely the bug the websocket transport was already fixed for — `wss._send_prepared_command` re-reads `_sock` under the lock with the comment "a check made before acquiring the lock can be stale by the time the send happens." BLE never got the same treatment.

## Fix

Mirror the wss fix on both paths: re-read under the lock, bind once, and finish the exchange on the bound client. On send, a `None` re-read raises `NotConnectedError`; a drop the backend notices mid-payload surfaces as bleak's own error on a dead client, never `AttributeError`. On receive, a `None` re-read returns, and a `BleakError` mid-read abandons the reply the way a truncated response already is — `_on_disconnected` separately fails the command in flight.

## Tests

Both fail against unpatched `main` (with the two `AttributeError`s above), verified:

- `test_a_drop_mid_send_raises_the_contract_error`
- `test_a_drop_mid_reply_abandons_the_read_cleanly`

870 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -15,6 +15,7 @@ from uuid import UUID
 
 import bleak_retry_connector
 from bleak import BleakClient, BleakGATTCharacteristic, BLEDevice
+from bleak.exc import BleakError
 from bleak_retry_connector import BleakClientWithServiceCache
 
 from .exceptions import NotConnectedError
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Re-reading `_ble_client` under the lock (rather than trusting the outer check) closes both windows in the send and receive paths. Bounding the receive loop's read to BleakError so a mid-reply drop is abandoned cleanly, instead of crashing the notification-dispatch task where nothing would have caught it, is the more important of the two fixes.

PR #581, opened 2026-08-04
Ignore a state frame that arrives before start()

trailro:
## What

`PitBoss.__init__` registers `_on_state_received` on the transport immediately, but `self.spec` — the thing that parses frames — only exists once `start()` has resolved it. A transport can already be connected and receiving in that window: Home Assistant connects through `reset_device` during discovery and calls `start()` later (the ordering `ble.py`'s own comments document). A frame arriving then raises:

```
AttributeError: 'PitBoss' object has no attribute 'spec'
```

— inside the transport's dispatch, where no caller sees it. On BLE that's an unhandled task exception in bleak's notification handler; on WSS the read loop's per-payload handler logs it.

## Fix

The pre-`start()` frame is ignored with a DEBUG log. There is nothing useful to do with it — no spec exists to parse it, and the next push arrives seconds after `start()` completes. (Resolving the spec in `__init__` instead would move where `InvalidGrill` raises, which downstream setup code sequences around deliberately, so the guard is the surgical fix.)

## Test

`test_state_arriving_before_start_is_ignored` — a frame before `start()` must not crash, and a frame after `start()` must parse as ever. Fails against unpatched `main` with the `AttributeError` above, verified.

886 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -174,6 +174,15 @@ class PitBoss:
             status_payload,
             temperatures_payload,
         )
+        if not hasattr(self, "spec"):
+            # The callback is registered in `__init__` but the spec only
+            # exists once `start()` has resolved it -- and a transport can
+            # already be connected and receiving before then: Home Assistant
+            # connects through `reset_device` during discovery and calls
+            # `start()` later. A frame in that window has nothing to parse
+            # it; the next push arrives seconds after `start()` completes.
+            _LOGGER.debug("Ignoring state received before start()")
+            return
         state = StateDict()
         if status_payload and (
             new_state := self.spec.control_board.parse_status(status_payload)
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Verified, merging.

**886 tests**, `ruff check`, `ruff format --check`, `mypy .` clean. Mutating the guard fails exactly the test you named:

```
if not hasattr(self, "spec"):  ->  if False:
    test_state_arriving_before_start_is_ignored   FAILS (AttributeError)
```

I checked the one thing the description does not cover: **`_on_vdata_received` has the same shape but not the same bug.** It never touches `self.spec` — it parses the payload itself and dispatches — so the guard belongs only where you put it, and there is no sibling fix waiting.

Agreed on not resolving the spec in `__init__`. `InvalidGrill` raising from the constructor rather than from `start()` would land in ha-pitboss's setup path, where the retry and reauth behaviour is sequenced around exactly which call raises. That is a much larger change than the bug warrants.

The reachability is the part that makes this worth having: ha-pitboss really does connect through `reset_device` during discovery and call `start()` afterwards, so the window is not theoretical, and a frame landing in it dies inside bleak's notification handler where nobody sees it.

PR #582, opened 2026-08-04
Fail in-flight commands on an explicit BLE disconnect

trailro:
## What

`BleConnection._disconnect_locked` never fails the commands in flight — that is left entirely to bleak invoking the disconnected callback. Two real paths skip it:

- `client.disconnect()` itself raises (caught and logged with "Bluetooth is awful" — accurate), so the backend may never deliver the callback;
- a backend that does not fire `disconnected_callback` for a locally-initiated disconnect.

Either way, an in-flight `send_command` waits out its full 30s timeout to learn what `disconnect()` already knew. Reproduced: command in flight, `disconnect()` with a client whose `disconnect` raises `BleakError` — the command was still pending after `disconnect()` returned.

The websocket transport already fails its own in-flight commands on wind-down; BLE now matches.

## Fix

One call at the end of `_disconnect_locked`: `await self._fail_pending_commands()`. Idempotent with the callback path — `_fail_pending_commands` snapshots-and-clears under the lock and skips already-done futures, so the callback arriving as well is harmless.

## Test

`test_disconnect_fails_the_commands_in_flight` — in-flight command, disconnect whose `client.disconnect()` raises; the caller must get `NotConnectedError` promptly. Fails against unpatched `main` (times out with the command still pending), verified.

886 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/ble.py
+++ b/pytboss/ble.py
@@ -176,6 +176,12 @@ class BleConnection(Transport):
             # `NotConnectedError` every other transport raises.
             self._ble_client = None
         self._is_connected = False
+        # Not left to bleak's disconnected callback: when `disconnect()`
+        # itself raised above, or on a backend that does not fire the
+        # callback for a locally-initiated disconnect, nothing else fails
+        # the commands in flight and their callers wait out the full
+        # timeout. Idempotent with the callback path.
+        await self._fail_pending_commands()
 
     async def reset_device(self, ble_device: BLEDevice):
         """Resets the BLE device used for transport.
--- a/tests/test_ble.py
+++ b/tests/test_ble.py
@@ -844,3 +844,38 @@ async def test_a_drop_mid_reply_abandons_the_read_cleanly():
 
     client.read_gatt_char = mock.AsyncMock(side_effect=clear_mid_read)
     await conn._on_rpc_data_received(None, bytearray(len(reply).to_bytes(4, "big")))
+
+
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
Verified, merging.

**886 tests**, `ruff check`, `ruff format --check`, `mypy .` clean. Mutating the fix fails exactly the test you named:

```
drop `await self._fail_pending_commands()` from _disconnect_locked
    test_disconnect_fails_the_commands_in_flight   FAILS (times out, command still pending)
```

**Checked the reentrancy, since that is what would make this fix worse than the bug.** `_disconnect_locked` runs under `_lifecycle_lock`; `_fail_pending_commands` takes `_lock`. Different locks, and `_disconnect_locked` holds no `_lock` of its own, so there is no path where this awaits a lock it already owns. Worth stating explicitly because `ble.py` reuses the base `_lock` for GATT serialisation, so the two are easy to conflate.

The idempotency argument holds too: `_fail_pending_commands` snapshots and clears `_rpc_futures` under the lock and skips futures that are already done, so the callback arriving afterwards finds nothing to do rather than raising `InvalidStateError`.

Matching the websocket transport here is the right call. "The link dropped" and "I disconnected" are different events, but they owe a caller the same thing, and the BLE path was leaving that entirely to a callback the backend is not obliged to deliver — with `disconnect()` itself already known to fail sometimes, which the "Bluetooth is awful" comment above it has been documenting for a while.

PR #583, opened 2026-08-04
Raise from a fire-and-forget send the other transports refuse

trailro:
## What

`send_command_without_answer` on the HTTP transport spawns `_send_and_forget`, whose `except Exception` deliberately swallows everything — the right call for "the board rebooted before it could answer". But the `NotConnectedError` from `_send_prepared_command`'s own `_started` guard is swallowed by the same net, so:

```python
conn = HttpConnection(host)          # never connected
await PitBoss(conn, model).reboot()  # returns cleanly -- nothing was sent
```

BLE and WSS raise `NotConnectedError` synchronously from the base-class `send_command_without_answer` on the identical call — the three transports disagree on the contract, and the HTTP one reports success for a send that never left.

## Fix

The connected-check runs synchronously in `send_command_without_answer`, before the task is spawned — same predicate `_send_prepared_command` uses. The background swallow keeps covering only what it was written for: in-flight endings from a board that is genuinely gone mid-request.

## Tests

- `test_fire_and_forget_on_a_dead_transport_says_so` — never-connected and connected-then-disconnected both raise. Fails against unpatched `main`, verified.
- `test_fire_and_forget_still_swallows_the_expected_silence` — a failing in-flight request stays a clean ending, pinning that the fix did not over-reach.

887 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/http.py
+++ b/pytboss/http.py
@@ -175,7 +175,15 @@ class HttpConnection(Transport):
         :param method: The method to call.
         :param params: Parameters to send with the command.
         :param timeout: Ignored. Nothing is awaited, so nothing can elapse.
+        :raise pytboss.exceptions.NotConnectedError: If the transport is not
+            connected. Raised here rather than swallowed in the background
+            task: "the board rebooted before answering" and "nothing was
+            ever sent" are different endings, and the other transports
+            raise for the second one -- a caller's `reboot()` should not
+            report success on a transport that was never connected.
         """
+        if not self._started or self._session is None or self._session.closed:
+            raise NotConnectedError("Not connected")
         cmd = await self._prepare_command(method, params)
         task = self._loop.create_task(self._send_and_forget(cmd))
         # Held so the task is not collected mid-flight, and so `disconnect()`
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Verified, merging.

**887 tests**, `ruff check`, `ruff format --check`, `mypy .` clean. Mutating the fix fails exactly the test you named:

```
drop the synchronous connected-check
    test_fire_and_forget_on_a_dead_transport_says_so   FAILS
```

**Confirmed the cross-transport claim rather than taking it from the description** — a never-connected instance of each, same call:

```
wss    -> NotConnectedError
http   -> NotConnectedError   (with this branch; returns cleanly without it)
```

So the three now agree, and they did not before.

The second test is the one that makes this safe to take. Narrowing a deliberate `except Exception` is exactly where a fix over-reaches, and `test_fire_and_forget_still_swallows_the_expected_silence` pins that the swallow still covers what it was written for — a board that goes away mid-request, which is the normal ending for `Sys.Reboot`. Without that test this change would be one edit away from turning every reboot into a raised exception.

This is the fourth defect found on the HTTP transport that I wrote and reviewed myself, and the pattern in all four is the same: the code was right for the transport I was thinking about and wrong for the contract the other two already kept.

PR #584, opened 2026-08-04
Make connect()'s own handshake cancellable, and roll back on failure

trailro:
## What

Two holes in `_connect_locked`, both proved by running code:

**1. A failed `connect()` leaks the owned session.** `_connect_locked` creates its `ClientSession`, then `_ws_connect` raises `GrillUnavailable` — nothing closes what was just opened. A caller that abandons the transport (a config flow constructing a fresh one per attempt) leaks one session per retry, each printing aiohttp's `Unclosed client session`. `http.connect()` has exactly the rollback this path lacked.

**2. `connect()`'s own handshake is the one the lifecycle work never covered.** #572 made the *reconnect loop's* handshake cancellable via `_connect_task`, but `connect()`'s own remained a bare `await self._ws_connect()` — executed while holding `_lifecycle_lock`. `disconnect()` blocks on that same lock, so against a host that accepts nothing, `disconnect()` queues behind the stalled handshake for aiohttp's session timeout (5 minutes by default). Measured: with `connect()` parked in the handshake, `disconnect()` had not returned after 0.5s and could only be freed by externally cancelling the connect task.

## Fix

- The handshake runs as `_connect_task`, exactly as the reconnect loop's does, reusing the `_handshake_cancelled` disambiguation #572 introduced.
- `disconnect()` cancels `_connect_task` *before* taking the lifecycle lock — the lock `connect()` is holding is precisely why waiting first cannot work.
- The cancelled `connect()` rolls back and raises `NotConnectedError("disconnect() was called while connecting")`; any other failure (including `GrillUnavailable`) rolls back — owned session closed and released — and re-raises. `_connect_task`/`_handshake_cancelled` are cleared in a `finally`, same discipline as the reconnect path.

## Tests

Both fail against unpatched `main`, verified:

- `test_a_failed_connect_does_not_leak_the_owned_session` — the failing run even prints the "Unclosed client session" warning this removes.
- `test_disconnect_interrupts_a_connect_stuck_in_its_handshake` — `disconnect()` completes within its 3s bound and `connect()` surfaces the documented error; on main the disconnect times out behind the lock.

887 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -93,10 +93,34 @@ class WebSocketConnection(Transport):
             if not self._session_owned:
                 raise NotConnectedError("The session given to this transport is closed")
             self._session = ClientSession(loop=self._loop)
-        self._sock = await self._ws_connect()
-        self._keep_running = True
-        self._stopping.clear()
-        self._subscribe_task = self._loop.create_task(self._subscribe())
+        # Run as a task for the same reason the reconnect loop runs its own:
+        # no flag reaches a handshake in flight. `disconnect()` cancels
+        # `_connect_task` *before* taking the lifecycle lock -- the lock this
+        # call is holding -- so a caller is not parked behind a stalled
+        # handshake for aiohttp's full session timeout.
+        self._connect_task = self._loop.create_task(self._ws_connect())
+        try:
+            self._sock = await self._connect_task
+            self._keep_running = True
+            self._stopping.clear()
+            self._subscribe_task = self._loop.create_task(self._subscribe())
+        except BaseException as ex:
+            # The rollback `http.connect()` already has: a failed or
+            # cancelled connect must not leak the session this call opened.
+            # A config flow constructing a fresh transport per attempt
+            # otherwise leaks one per retry.
+            if self._session_owned and self._session is not None:
+                if not self._session.closed:
+                    await self._session.close()
+                self._session = None
+            if isinstance(ex, asyncio.CancelledError) and self._handshake_cancelled:
+                raise NotConnectedError(
+                    "disconnect() was called while connecting"
+                ) from ex
+            raise
+        finally:
+            self._connect_task = None
+            self._handshake_cancelled = False
         await self._subscribed.wait()
 
     def _check_not_reentrant(self) -> None:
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Both fixes are right, and the second one closes the hole I left open when I described the design on #567 — I only ever thought about the reconnect loop's handshake, not `connect()`'s own. Cancelling before taking the lock, rather than after, is the part I would have got wrong: waiting for a lock that the stalled call is holding is not waiting, it is deadlock with a timeout.

**887 tests**, `ruff check`, `ruff format --check`, `mypy .` clean.

I mutated each mechanism separately:

```
drop the session rollback           -> test_a_failed_connect_does_not_leak_the_owned_session  FAILS
drop the pre-lock cancel            -> test_disconnect_interrupts_a_connect_stuck_in_...      FAILS
move the cancel INSIDE the lock     -> test_disconnect_interrupts_a_connect_stuck_in_...      FAILS
convert ANY cancel, not just ours   -> 34 passed
```

That third one is worth having: it pins the *ordering*, not just the presence of the cancel, so a later reader cannot tidy it inside the `async with` and keep green.

## The `_handshake_cancelled` guard is load-bearing and nothing pins it

This is coverage rather than a defect — the code is correct as written. I probed it before saying so, since a missing test is only worth raising if the path is reachable, and this one is: any caller that bounds its own `connect()`. Nothing in the scenario calls `disconnect()`.

```
as-written       TimeoutError  <- caller's deadline honoured          after 1.00s
guard-removed    NotConnectedError: disconnect() was called while connecting
```

Without the `and self._handshake_cancelled`, a caller's `asyncio.timeout` is swallowed and replaced with a message that is simply untrue. That is the same shape as the defect on #572 — a bound the caller asked for, silently not honoured — one level up, in the path this PR adds.

Both assertions are observable outcomes rather than implementation details: `TimeoutError` reaching the caller, and the error not being the disconnect one. The scenario above is the whole test; no counter mock needed.

**Not holding the merge for it.** The guard is present and correct, this is a follow-up, and I would rather land two real fixes than trade them for a test.

## One thing I checked and am satisfied with

`_handshake_cancelled` cannot be left stale. `disconnect()` reads `_connect_task` and sets the flag in the same synchronous block with no `await` between them, so the owning coroutine's `finally` cannot clear it in the gap — there is no gap. Both owners (`_connect_locked` and `_subscribe`) clear it in a `finally`, so whichever holds the in-flight handshake resets it.

PR #585, opened 2026-08-04
Deliver callbacks from their own task, not from the read loop

trailro:
> **Stacked on #584** (same file, same lifecycle machinery) — the first commit here is that PR; this one is `deliver callbacks from their own task, not from the read loop`. Merging #584 first collapses this to a one-commit diff.

## What

The websocket read loop awaits `_handle_message` — and with it every subscriber callback — inline. A subscriber that issues an RPC from its callback therefore waits on a reply that only the read loop it is blocking can deliver. Measured: a `ping()` that completes in 0.000s outside a callback times out at exactly its full deadline inside one, with the answer sitting unread in the socket the whole time — and every state update queues behind it (a frame sent at t=0.2s was delivered at t=2.0s). At the default timeout that is a 30s stall of all state updates per callback-issued RPC. A slow callback also holds up `disconnect()`, which awaits the subscribe task.

The API layer explicitly blesses subscriber callbacks; its own `_callback_lock` comment documents dispatching outside the lock precisely so subscribers can call back into the API — which then doesn't work on this transport for any command that expects an answer.

## Fix

Callbacks flow through a queue consumed by a dedicated dispatch task:

- the read loop only enqueues (`put_nowait`) — it never awaits user code again, so replies are processed the moment they arrive;
- a **single** consumer preserves delivery order (arrival order), which per-frame `create_task` would not guarantee;
- the queue is rebuilt on each `connect()` so a stale frame from a previous session is never delivered into a new one;
- wind-down cancels the dispatcher rather than draining it — draining means awaiting the very user code this task exists to keep out of the way;
- the reentrancy guard covers the dispatch task too: a callback calling `disconnect()` would be a lifecycle call cancelling its own caller, so it is refused with the same error as from the read loop.

`test_state_callback_may_send_commands`' docstring — which documented "awaiting the response inside a callback can never complete" — is updated: it now can.

## Tests

- `test_a_callback_can_await_a_full_round_trip` — RPC from a state callback completes with the real result in <2s. Fails against the base branch (times out), verified.
- `test_an_outside_disconnect_is_not_blocked_by_a_parked_callback` — a subscriber sleeping forever no longer holds up `disconnect()`. Fails against the base branch, verified.
- `test_a_callback_calling_disconnect_is_told_no` — the reentrancy contract, now exercised through the dispatch task.
- The two direct `_handle_message` vdata tests drive the dispatcher explicitly via a small delivery helper (`task_done`/`join`, deterministic).

889 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -2,7 +2,7 @@
 
 import asyncio
 import logging
-from asyncio import AbstractEventLoop, Event, Lock, Task
+from asyncio import AbstractEventLoop, Event, Lock, Queue, Task
 from contextlib import suppress
 from typing import Any
 from uuid import uuid4
@@ -58,6 +58,14 @@ class WebSocketConnection(Transport):
         self._url = f"{base_url}/to/{grill_id}"
         self._app_id = app_id or str(uuid4()).split("-")[-1]
         self._subscribe_task: Task | None = None
+        # Callbacks are delivered from their own task, fed by this queue, so
+        # the read loop never awaits user code. A callback that issues an
+        # RPC otherwise waits on a reply only the read loop it is blocking
+        # can deliver -- burning the command's whole timeout -- and every
+        # state update queues behind it. Rebuilt on each connect so a stale
+        # frame from a previous session is never delivered into a new one.
+        self._callback_queue: Queue[tuple[str, Any]] = Queue()
+        self._dispatch_task: Task | None = None
         self._subscribed = Event()
         self._stopping = Event()
         self._keep_running = False
@@ -103,7 +111,9 @@ class WebSocketConnection(Transport):
             self._sock = await self._connect_task
             self._keep_running = True
             self._stopping.clear()
+            self._callback_queue = Queue()
             self._subscribe_task = self._loop.create_task(self._subscribe())
+            self._dispatch_task = self._loop.create_task(self._dispatch_callbacks())
         except BaseException as ex:
             # The rollback `http.connect()` already has: a failed or
             # cancelled connect must not leak the session this call opened.
@@ -132,7 +142,10 @@ class WebSocketConnection(Transport):
         lock awaits the very subscribe task this caller is running on -- a
         deadlock the guard exists to prevent, not cause.
         """
-        if self._subscribe_task is asyncio.current_task():
+        current = asyncio.current_task()
+        if current is not None and (
+            current is self._subscribe_task or current is self._dispatch_task
+        ):
             raise RuntimeError(
                 "Cannot stop this transport from a callback it is dispatching"
             )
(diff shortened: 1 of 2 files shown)


trailro:
Rebased onto `main` — with #584 in, the stack collapsed to the single dispatcher commit as promised. 893 pass, ruff/format/mypy clean.

dknowles2 (approved):
This is the good kind of fix: the bug is a design consequence rather than a slip, and the fix removes the category instead of the instance. Awaiting user code from the read loop meant any callback that issued an RPC deadlocked against the only task that could answer it — and the API layer's own `_callback_lock` comment explicitly promises subscribers may call back in.

**893 tests** (up from your 889 — the rebase onto #584 brought its two along), `ruff check`, `ruff format --check`, `mypy .` clean.

Mutated each mechanism:

```
await the status callback inline again  -> test_a_callback_can_await_a_full_round_trip       FAILS
                                           test_an_outside_disconnect_is_not_blocked_by_...  FAILS
                                           test_a_temperatures_only_push_reaches_the_...     FAILS
drop the dispatch task from the guard   -> test_a_callback_calling_disconnect_is_told_no     FAILS
drain instead of cancel                 -> test_an_outside_disconnect_is_not_blocked_by_...  FAILS
reuse the queue across connects         -> 36 passed
```

## The queue rebuild is load-bearing and unpinned

Coverage, not a defect — the line is there and correct. I probed it before calling it a gap, because a missing test only matters if the path is reachable, and this one is *more* reachable after this PR than before it: the whole point is that callbacks may now do slow things, which is exactly what leaves frames queued at teardown.

Frames arriving faster than a slow subscriber drains them, torn down mid-backlog, transport reused:

```
with-rebuild     delivered into session 2 = ['session-2-frame']                       stale=0
without-rebuild  delivered into session 2 = ['session-1-frame-B', 'session-2-frame']  stale=1
```

The stale frame is delivered into the new session *ahead of* the live one, so a subscriber's first update after a reconnect is a temperature reading from before the drop. Observable, and the assertion is just "no session-1 payload reaches a session-2 subscriber" — no timing involved.

**Not holding the merge.** The other three mechanisms are pinned, this one is present and correct, and I would rather land the fix.

## Two small things, neither blocking

`Queue()` is unbounded. With a 5-second push cadence and a callback that can now legitimately take a command timeout, a backlog is possible where it was not before — the read loop no longer applies backpressure, which is the point, but it also means nothing bounds the queue. Probably fine forever; worth knowing it is unbounded by choice rather than by oversight.

`_dispatch_callbacks` catches `Exception`, so a callback that raises `CancelledError` for its own reasons kills the dispatcher silently and no further callbacks are delivered for the life of the connection. Narrow — it needs a subscriber that cancels itself — but the failure is silent and permanent, where every other callback error is logged and survived.

## What I checked and am satisfied with

The queue swap in `connect()` cannot strand the old dispatcher on the wrong queue: `_connect_locked` calls `_stop_subscribing()` first, which cancels and awaits `_dispatch_task` before the rebuild happens, so there is no window where a live dispatcher holds an item from a queue that has been replaced under it. That would have been a `task_done() called too many times`, and it is not reachable.

Single consumer rather than a task per frame is the right call, and worth having said explicitly — arrival order is the only order a subscriber can reason about, and `create_task` per frame would have quietly stopped guaranteeing it the moment any callback awaited anything.

PR #586, opened 2026-08-04
Answer "not served" the way the docstrings promise

trailro:
## What

Three wrappers document a graceful answer for a device that does not serve their RPC:

- `WiFi.scan()` — "Empty when the device does not serve it"
- `PitBoss.get_loader_version()` — "`None` when the grill does not serve `PBL.GetLoaderVersion`"
- `PitBoss.scan_wifi_networks()` — "Empty if … the grill serves no loader"

None of that was true. An unserved method is answered with an error frame, which `send_command` raises as `RPCError` — the `isinstance` guards only cover a *successful* reply of the wrong shape. The tests certified the docstrings by mocking replies the "does not serve it" case cannot produce (`send_command` returning `None`).

## Why the fix is a fix and not a doc change

Verified against a live grill: an unknown method answers

```json
{"code": 404, "message": "No handler for Totally.Bogus"}
```

while a **failing served** method answers a different code (e.g. `400` for a bad `FS.Get`). So "not served" is genuinely distinguishable from "went wrong", and the documented graceful answers can be real: the three wrappers catch `RPCError` with the new `METHOD_NOT_FOUND_CODE = 404` (documented next to `UNAUTHORIZED_CODE`, with the live-grill evidence) and return their documented empty/`None`. **Any other code keeps raising** — a loader that exists but fails is not silently read as absent, pinned by tests.

`FakeTransport` now models the 404 frame the way the firmware sends it (unknown method → 404, failing handler → its own code), so the fixture can no longer mask this class of bug.

## Tests

The three rewritten "does not serve it" tests drive the real error path and fail against unpatched `main` (verified); new failure-path tests pin that other codes still raise; `test_other_rpc_errors_keep_their_code` now distinguishes the unknown-method 404 from a served handler's own code.

894 pass, `ruff check`, `ruff format --check` and `mypy .` clean. No public signatures changed; ha-pitboss uses none of these three wrappers.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -11,11 +11,11 @@ from typing import Any
 
 from .codec import WIFI_KEY, encode, timed_key
 from .config import Config
-from .exceptions import UnsupportedOperation
+from .exceptions import RPCError, UnsupportedOperation
 from .fs import FileSystem
 from .grills import Grill, StateDict, get_grill
 from .ota import OTA
-from .transport import RPCResult, Transport, as_dict
+from .transport import METHOD_NOT_FOUND_CODE, RPCResult, Transport, as_dict
 from .wifi import WiFi
 
 _UPTIME_TTL = 60.0
@@ -684,10 +684,17 @@ class PitBoss:
         `get_firmware_version()` reports: a grill on firmware 0.5.7 answers
         `0.2.2` here.
 
-        `None` when the grill does not serve `PBL.GetLoaderVersion`, which
-        `list_rpcs()` reports without a round trip that errors.
+        `None` when the grill does not serve `PBL.GetLoaderVersion`.
         """
-        result = await self._conn.send_command("PBL.GetLoaderVersion", {})
+        try:
+            result = await self._conn.send_command("PBL.GetLoaderVersion", {})
+        except RPCError as ex:
+            if ex.code == METHOD_NOT_FOUND_CODE:
+                # The documented None. Without this, a grill without the
+                # loader RPC raised instead of answering what the docstring
+                # promises.
+                return None
+            raise
         if isinstance(result, dict):
             return result.get("loaderVersion")
         return None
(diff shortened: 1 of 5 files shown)


dknowles2 (approved):
This is my bug, and the way it survived is worth naming.

I wrote those three docstrings in #573 and wrote tests that mocked `send_command` returning `None` — a shape the "does not serve it" case **cannot produce**. A device that does not serve a method answers an error frame, not a result-less reply. So the tests certified the docstring rather than the behaviour, and the `isinstance` guards I was so pleased with only ever covered a *successful* reply of the wrong shape.

The tell was in my own prose. `WiFi.scan()` said `list_rpcs()` was "worth checking first, since a round trip that errors" reads the same as an empty list — I wrote a workaround for the bug into the docstring instead of noticing it was a bug. Your rewrite of that paragraph is better because it now describes a real distinction (empty vs. no networks) rather than papering over a raise.

That is the third time this pattern has landed: a mock producing a reply the real failure cannot send, agreeing with a docstring describing the same fiction. The other two were `OTA.get_status`'s invented shape and `OTA.Start` itself. Modelling the 404 frame in `FakeTransport` is the fix for the class, not just these three — the fixture can no longer answer `-1` for a method the firmware would 404.

**894 tests**, `ruff check`, `ruff format --check`, `mypy .` clean.

```
METHOD_NOT_FOUND_CODE 404 -> 405   3 tests FAIL
swallow every RPCError (wifi)      test_a_failing_scan_is_...        FAILS
swallow every RPCError (loader)    test_get_loader_version_...       FAILS
```

The second and third are the ones that matter: they pin that a **served method that fails** still raises. Catching `RPCError` and returning the documented empty unconditionally is the obvious over-reach here, and it would turn a loader that exists but is broken into one that silently does not exist. Good that it fails loudly.

Distinguishing 404 from a served handler's own code is what makes the documented behaviour honest rather than a guess, and putting the live-grill evidence next to the constant is the right place for it — that is the fact a future reader will want and cannot re-derive from the code.

Only guarding the first call in `scan_wifi_networks()` is right: once `PBL.StartWifiScan` is served, `PBL.GetWifiScanStatus` is too. They ship in the same loader.

PR #587, opened 2026-08-04
Deliver the hashability frozen=True advertises

trailro:
## What

`Grill` and `ControlBoard` are `frozen=True` dataclasses, which advertises hashability — but the generated `__hash__` hashes every field, and `commands`/`json` are dicts, `temp_increments`/`celsius_temp_increments` lists. So:

```python
>>> hash(get_grill("PB1600PS1"))
TypeError: unhashable type: 'dict'
>>> set(get_grills())
TypeError: cannot use 'pytboss.grills.Grill' as a set element
```

Deduplicating models via a set is the obvious thing a consumer does with a frozen type, and it fails at runtime.

## Fix

The container fields are excluded from the hash with `field(hash=False)`. Equality is untouched — it still compares every field — and since equal objects agree on the hashable subset, the `a == b ⟹ hash(a) == hash(b)` contract holds (objects differing only in a container field hash together and compare unequal, which is an allowed collision, not a violation).

No public signatures change; nothing downstream is affected — this only makes previously-crashing operations work.

## Test

`test_frozen_types_are_hashable` — every catalogue model hashes, none collide away in a set, and two `get_grill()` results for the same model are equal, hash equal, and dedupe to one. Fails against unpatched `main` with the `TypeError` above, verified.

892 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -297,8 +297,13 @@ class ControlBoard:
     name: str
     """Name of the control board."""
 
-    commands: dict[str, Command]
-    """Controller commands indexed by their slug."""
+    commands: dict[str, Command] = field(hash=False)
+    """Controller commands indexed by their slug.
+
+    Excluded from the hash (dicts are unhashable), so `frozen=True` delivers
+    the hashability it advertises. Equality still compares every field, and
+    equal boards agree on the hashable subset, so the `a == b` implies
+    `hash(a) == hash(b)` contract holds."""
 
     _status_js_func: str | None
     """JavaScript function body that parses a status reply."""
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Verified, merging. **892 tests**, `ruff check`, `ruff format --check`, `mypy .` clean.

```
drop hash=False from json       -> test_frozen_types_are_hashable   FAILS
drop hash=False from commands   -> test_frozen_types_are_hashable   FAILS
```

Both container fields are pinned separately, which matters — one of them alone would still leave the type unhashable, and a single test asserting "it hashes" could have been satisfied by either fix in isolation.

## The two things I check on `grills.py` by default

**Constructor order is untouched.** These are `frozen=True` but not `kw_only`, so field order is public API and positional callers break silently if it moves. `field(hash=False)` on a field that already had no default adds no default, so nothing reordered:

```
Grill:        name, control_board, has_lights, has_mpc, min_temp, max_temp,
              meat_probes, temp_increments, json, celsius_temp_increments
ControlBoard: name, commands, _status_js_func, _temperatures_js_func
Command:      slug, _hex, _js_func
```

**The hash/eq contract holds across the real catalogue**, not just the sampled models in the test:

```
137 models  ->  137 distinct hashes  ->  set() of size 137
PB1600PS1 / PBV4PS2 / PB850DX: equal=True, hash equal=True, dedupe=1
contract violations: 0
```

No collisions at all in practice, so the theoretical "differ only in a container field" collision you allow for does not actually occur in the shipped data — the hashable subset is discriminating enough on its own. Worth knowing, since a set of every model is precisely the operation this enables.

Your reasoning about the contract is the right way round, and worth restating because it is the part people get backwards: excluding fields from the hash is safe, excluding them from `__eq__` would not be. Equal objects still agree on the hashable subset, so `a == b ⟹ hash(a) == hash(b)` survives; the reverse implication was never promised.

PR #588, opened 2026-08-04
Give the tests that assert nothing something to fail on

trailro:
## What

Five test-hygiene fixes, all found by collection sweep and mutation testing. None change library code — this PR makes existing coverage honest.

**1. Two tests that never ran.** `TestController.parse_status`/`parse_temperatures` lack the `test_` prefix, so pytest collects nothing from that class — verified with `--collect-only`. They are the only direct unit tests of the `ControlBoard` parsers, and they have passed-if-run for as long as they've existed. Renamed; the suite gains two genuinely running tests.

**2. A test that asserts nothing about the code under test.** `test_set_wifi_credentials_is_not_the_config_service_key`'s load-bearing assertion exercises `codec.py` alone — the API call above it never feeds it. Proved by mutation: switching `set_wifi_credentials` to encode under the *grill* key still passed one parametrized variant. The sibling `test_set_wifi_credentials_obfuscates_the_password` already pins the key end-to-end (its docstring even explains why the round trip is the right assertion), so the inert test is deleted as false coverage rather than duplicated.

**3. A test satisfied by the code it was meant to check.** `test_disconnect_stops_a_request_still_in_flight` asserted `_pending == set()` — which `disconnect()`'s own `clear()` satisfies whether or not anything was cancelled. Proved by mutation: removing the `task.cancel()` entirely still passed. The request is now genuinely parked server-side (an event confirms it is on the wire) and the task's cancellation is observed directly. The mutation now fails it.

**4. A mutation survivor in the reconnect loop.** The `finally` clearing `_handshake_cancelled` had no test: removing it survived the whole wss suite. Its scenario — a cancel that never landed (the handshake finished first) must not make the *next*, caller-owned cancellation look like ours and be swallowed — is now pinned; the mutation fails on the intended assert.

**5. Unspecced mocks.** The ota/wifi fixtures used bare `AsyncMock()` while the fs/config siblings use `AsyncMock(spec=Transport)` — a call to a nonexistent transport method was silently absorbed. Specced to match.

892 pass (890 previously + the two recovered, − the one deleted), `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -10,7 +10,7 @@ import pytest
 from freezegun import freeze_time
 
 from pytboss import api, grills
-from pytboss.codec import WIFI_KEY, decode, encode, timed_key
+from pytboss.codec import WIFI_KEY, decode, timed_key
 from pytboss.exceptions import (
     InvalidGrill,
     RPCError,
@@ -1188,16 +1188,6 @@ async def test_set_wifi_credentials_does_not_send_the_password_in_the_clear(
     with mock.patch.object(conn, "_send_prepared_command", capture):
         await pitboss.set_wifi_credentials("Kitchen", "hunter2")
     assert "hunter2" not in json.dumps(sent)
-
-
-async def test_set_wifi_credentials_is_not_the_config_service_key(
-    pitboss: api.PitBoss, conn: FakeTransport
-):
-    """Encoding under the grill password key must not decode as the WiFi one."""
-    await pitboss.set_wifi_credentials("Kitchen", "hunter2")
-    assert conn.wifi_credentials is not None
-    wrong = decode(bytes.fromhex(encode(b"hunter2").hex()), key=WIFI_KEY)
-    assert wrong != b"hunter2"
 
 
 async def test_debug_pstate(pitboss: api.PitBoss, conn: FakeTransport):
--- a/tests/test_grills.py
+++ b/tests/test_grills.py
@@ -57,11 +57,13 @@ class TestCommand:
 
 
 class TestController:
-    def parse_status(self):
+    # `test_` prefixes, or pytest collects nothing from this class: both
+    # bodies passed for as long as they existed, but neither ever ran.
+    def test_parse_status(self):
         ctrl = grills_lib.ControlBoard("PBx", {}, "return {'foo': message}", None)
         assert ctrl.parse_status("bar") == {"foo": "bar"}
 
-    def parse_temperatures(self):
+    def test_parse_temperatures(self):
         ctrl = grills_lib.ControlBoard("PBx", {}, "", "return {'foo': message}")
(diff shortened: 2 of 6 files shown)


trailro:
Rebased onto `main` — the wifi import conflict (#586) and the tail of `test_wss.py` (#585) both resolved by keeping both sides. 898 pass, ruff/format/mypy clean.

dknowles2 (approved):
Verified the strongest way available — each mutation run against **both** branches, so the claim "this was uncovered" is demonstrated rather than asserted:

```
                                          main        this PR
drop the pending-task cancel (item 3)     SURVIVES    CAUGHT  test_disconnect_stops_a_request_still_in_flight
drop the _handshake_cancelled reset (4)   SURVIVES    CAUGHT  test_a_cancel_that_never_landed_...
control: wrong wifi codec key             CAUGHT      CAUGHT  test_set_wifi_credentials_obfuscates_...
```

The control matters for item 2: after deleting the inert test the key is **still** pinned, so the deletion removed false coverage rather than real coverage. That is the thing to check before deleting a test, and it holds.

`TestController` collects **0 items** on `main` — confirmed with `--collect-only`. Two parser tests that have never run, ever.

**898 tests** on the branch (897 on current `main`, +2 recovered, −1 deleted), `ruff check`, `ruff format --check`, `mypy .` clean.

## The deleted test is mine, and it is the fourth of a kind

I wrote `test_set_wifi_credentials_is_not_the_config_service_key` in #573. Its load-bearing line was

```python
wrong = decode(bytes.fromhex(encode(b"hunter2").hex()), key=WIFI_KEY)
assert wrong != b"hunter2"
```

which exercises `codec.py` and nothing else. The `await pitboss.set_wifi_credentials(...)` above it feeds nothing into the assertion — I wrote a test *about* the code instead of *against* it, and gave it a name that claimed otherwise.

That is the same failure as #586 an hour ago, and as `OTA.get_status` before that: an assertion that agrees with my belief rather than with the code. The pattern is specific enough to name now — **when I write a test to demonstrate a distinction, I tend to demonstrate the distinction rather than test the code that depends on it.**

Items 3 and 4 are the same shape from a different direction: assertions satisfied by something other than the mechanism under test. `_pending == set()` is true because `disconnect()` calls `clear()`, cancel or no cancel. Parking the request genuinely on the wire and observing the task's own cancellation is what makes it a test.

## One note

Specced mocks (item 5) is the quiet one but probably the highest-value line here: a bare `AsyncMock()` absorbs a call to a transport method that does not exist, so a wrapper calling a misspelled RPC helper would pass. `spec=Transport` on the ota/wifi fixtures brings them in line with the fs/config siblings, and the reason the others already had it is presumably that someone hit this before.

No library code changed, so nothing downstream moves.

PR #589, opened 2026-08-04
Isolate state and vdata subscribers from each other

trailro:
## What

`PitBoss._on_state_received` and `_on_vdata_received` fan out to subscribers with a bare `for cb in callbacks: await _invoke(...)` and no per-callback isolation. So the first subscriber to raise:

1. skips every subscriber registered after it for that update, and
2. lets the exception escape out of the receive path — on BLE that surfaces as an unhandled-task traceback inside bleak's notification dispatch.

Reproduced (two state subscribers, first raises): only the raiser ran; the second never saw the update, and `RuntimeError` propagated out of `_on_state_received`.

The transport layer's own `wss._dispatch_callbacks` already wraps its single callback in `except Exception` — but the multi-subscriber loop lives one layer up, in `PitBoss`, where it had no such guard.

## Fix

Each callback invocation is wrapped in `try/except Exception` with `_LOGGER.exception`, matching the dispatcher's discipline. One raising subscriber no longer starves the rest or escapes the receive path. (The cache is advanced before dispatch, as before, so there's no state corruption — the bug was purely a fan-out isolation gap.)

## Tests

`test_one_raising_state_subscriber_does_not_starve_the_rest` and the vdata sibling — a raising subscriber must not skip the later ones or raise out. Both fail against unpatched `main`, verified.

### DIFF
--- a/pytboss/api.py
+++ b/pytboss/api.py
@@ -211,7 +211,15 @@ class PitBoss:
             # TODO: Run callbacks concurrently
             # TODO: Send copies of state so subscribers can't modify it
             for callback in callbacks:
-                await _invoke(callback, self._state)
+                try:
+                    await _invoke(callback, self._state)
+                except Exception:
+                    # Isolated per subscriber: without this the first one to
+                    # raise skips every subscriber registered after it for
+                    # that update, and the exception escapes into the
+                    # transport's dispatch -- on BLE, an unhandled task
+                    # traceback in bleak's notification handler.
+                    _LOGGER.exception("Error in state subscriber")
 
     async def _on_vdata_received(self, payload: str | dict):
         # Both, because the transports differ in what they can hand over.
@@ -236,7 +244,11 @@ class PitBoss:
             # TODO: Run callbacks concurrently
             # TODO: Send copies of state so subscribers can't modify it
             for callback in callbacks:
-                await _invoke(callback, vdata)
+                try:
+                    await _invoke(callback, vdata)
+                except Exception:
+                    # Isolated per subscriber, as in `_on_state_received`.
+                    _LOGGER.exception("Error in vdata subscriber")
 
     async def _authenticate(self, params: dict) -> dict:
         """Return `params` with the encoded password added.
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Right fix, and the framing is right too: the transport dispatcher already had this discipline, and the multi-subscriber loop one layer up did not — which is exactly where it matters more, since a single callback failing is one thing and starving every subscriber registered after it is another.

```
drop state isolation   -> test_one_raising_state_subscriber_does_not_starve_the_rest  FAILS
drop vdata isolation   -> test_one_raising_vdata_subscriber_...                       FAILS
```

## `except Exception` rather than `BaseException` is load-bearing, and nothing pins it

Coverage, not a defect — you chose correctly. I probed rather than reasoned, because this is the third time this session the same class has come up and I would rather have the numbers:

```
as-written     CancelledError     <- cancel honoured   later subscriber ran=False
BaseException  returned normally  <- cancel SWALLOWED  later subscriber ran=True
```

Cancelling the dispatching task while a subscriber is parked: as written the cancellation lands and the remaining subscribers are correctly skipped. Widened to `BaseException`, the cancel is absorbed, dispatch carries on to the next subscriber, and the task returns as though nothing happened — so `wss._stop_subscribing`'s `await self._dispatch_task` would sit through however long the remaining subscribers take, having asked it to stop.

Reachable through the normal wind-down path, since `_dispatch_callbacks` is cancelled on disconnect and this loop is what it is awaiting.

## Worth one test rather than three

This is now the third unpinned "we deliberately do not swallow cancellation" guard: #584's `and self._handshake_cancelled`, #585's queue rebuild, and this. All three are correct, none is pinned, and all three fail the same way — a caller's cancellation silently becoming success.

Rather than three follow-up tests, one shared assertion would cover the class: **cancel the dispatch path mid-callback and assert `CancelledError` reaches the awaiter.** It exercises the wss dispatcher, this loop, and the connect-time guard in a single scenario, because they are all on the same path. Happy to write it if you would rather not — it is a test for code that is already correct, so there is no hurry either way.

**Not holding the merge.** Both isolation mechanisms are pinned and the fix is right.

PR #590, opened 2026-08-04
Survive a subscribe task cancelled from outside

trailro:
## What

`_stop_subscribing`'s `await self._subscribe_task` re-raises whatever the task ended with. The `except Exception` there (from the dead-task-poisoning fix) catches a task that *died* — but not one that ended **cancelled**: `CancelledError` is a `BaseException`. An outside cancellation of the subscribe task (a shutdown sweep, a test harness reaping lingering tasks) therefore:

1. escaped `disconnect()` into a caller that never asked to be cancelled (`current_task().cancelling() == 0`);
2. skipped the dispatch-task cancel below it — orphaning the dispatcher;
3. skipped the owned-session close in `_disconnect_locked` — leaking the `ClientSession`.

Reproduced end-to-end against a real aiohttp ws server: `disconnect()` raised `CancelledError`, owned session left open, dispatch task not done.

## Fix

A `CancelledError` from awaiting the dead subscribe task is absorbed (logged at DEBUG) so the wind-down completes — session closed, dispatcher cancelled, state cleared — **except** when this task is itself being cancelled (`asyncio.current_task().cancelling() > 0`), which must still propagate so a genuine cancellation of the caller is honoured. This is the one distinction that separates "the task I was waiting on ended cancelled" from "someone is cancelling me."

## Tests

- `test_disconnect_survives_an_externally_cancelled_subscribe_task` — owned session closed, dispatch task done, no `CancelledError` escapes. Fails against unpatched `main` (the reproduction), verified.
- `test_a_cancellation_aimed_at_the_caller_is_still_honoured` — a cancellation targeting the task running `disconnect()` still propagates; guards the fix against over-swallowing.

### DIFF
--- a/pytboss/wss.py
+++ b/pytboss/wss.py
@@ -173,6 +173,18 @@ class WebSocketConnection(Transport):
             # the lines below -- clearing state, and in `disconnect()`'s
             # case closing an owned session -- never ran.
             await self._subscribe_task
+        except asyncio.CancelledError:
+            # The subscribe task ended *cancelled* -- an outside cancellation
+            # (a shutdown sweep, a test harness reaping tasks), not our own
+            # caller being cancelled. `except Exception` does not catch this,
+            # so it used to escape `disconnect()` into a caller that never
+            # asked to be cancelled, and skipped the dispatch cancel and the
+            # owned-session close below. Distinguished from a cancellation
+            # aimed at *this* task, which must still be honoured.
+            task = asyncio.current_task()
+            if task is not None and task.cancelling() > 0:
+                raise
+            _LOGGER.debug("Subscribe task was cancelled")
         except Exception:
             _LOGGER.exception("Subscribe task ended with an unhandled error")
         finally:
--- a/tests/test_wss.py
+++ b/tests/test_wss.py
@@ -901,3 +901,69 @@ async def test_a_cancel_that_never_landed_does_not_mark_the_next_one(
 
     assert conn._handshake_cancelled is False
     await conn.disconnect()
+
+
+async def test_disconnect_survives_an_externally_cancelled_subscribe_task(
+    fake_server: TestServer,
+) -> None:
+    """An outside cancellation of the subscribe task must not break disconnect().
+
+    A shutdown sweep or a test harness can cancel the subscribe task from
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
`task.cancelling() > 0` is the right instrument, and it is the one I did not name when I raised this class on #589 — I described the distinction ("the task I was waiting on ended cancelled" vs "someone is cancelling me") without knowing there was a first-class way to ask. There is, and this is it.

Mutated in both directions, which is the part that matters here:

```
remove the handler entirely              -> test_disconnect_survives_an_externally_cancelled_...  FAILS
always re-raise (never absorb)           -> test_disconnect_survives_an_externally_cancelled_...  FAILS
never re-raise (drop the cancelling check)-> test_a_cancellation_aimed_at_the_caller_is_still_...  FAILS
```

A guard that absorbs cancellation is only safe if something pins that it *stops* absorbing at the right moment, and both halves fail independently here. That is a better job than I did asking for it — I suggested one test for the class; two tests pinning the two directions of one decision is the stronger shape, because the failure modes are opposites and a single test can only catch one of them.

## Where this leaves the pattern

This closes the third of the three unpinned cancellation guards I flagged today, and it is the one that turned out to be an actual defect rather than a coverage gap — `except Exception` genuinely does not catch `CancelledError`, so the wind-down really was being skipped, session and dispatcher and all.

Still open, both coverage rather than defects:

- **#584** — `and self._handshake_cancelled` in `_connect_locked`. Removing it converts a caller's `asyncio.timeout` into `NotConnectedError("disconnect() was called while connecting")`, which is untrue and silently swallows the caller's bound.
- **#585** — the `_callback_queue = Queue()` rebuild in `connect()`. Removing it delivers a previous session's frame into the new one, ahead of the live frame.

Both are one scenario each and neither needs a counter mock. No hurry — the code is correct in both.

## One thing I checked

Absorbing without `uncancel()` is correct here: the current task was never cancelled, so there is no cancellation request to retract. `uncancel()` would only be needed if this code had itself requested the cancellation it is now swallowing, which it has not — that path is `_stop_subscribing`'s own `_connect_task.cancel()`, and it is handled separately.

PR #591, opened 2026-08-04
Reject a truncated frame instead of parsing it into a falsy state

trailro:
## What

Every board's `status_function`/`temperature_function` opens with `if (parts < N) return null` — meant to drop a frame too short to read. But `parts` is the Array returned by `parseHexMessage`, and `Array < Number` coerces via the array's string form to `NaN < N`, which is **always false**. On **14 of 20 boards** the guard never fires (only PBA, PBB, PBD, PBE, PBT, PBX1 write `parts.length < N`).

With the guard dead, a truncated frame indexes past its end — every `parts[i]` is `undefined`, so every `=== 1` is false and every `convertTemperature` is null — and the routine returns a **fully-populated, entirely-falsy** status dict. `_on_state_received` sees that as truthy, merges it into the cache (flipping `moduleIsOn` and every error flag — `noPellets`, `highTempErr`, `motorErr` — to False), and pushes it to subscribers. Reproduced end-to-end on PBL: a 6-byte `FE0B00000000` fragment yielded `moduleIsOn: False` and clobbered a cached `True`.

**Reachability:** BLE. `_on_debug_log_received` processes one notification at a time with no reassembly, and only validates the firmware's `[len]` checksum when the payload ends in `]` — a headed fragment split across notifications lacks the trailing `]`, so the one integrity check is bypassed exactly on truncation. HTTP/WSS deliver fully-reassembled frames, so they are not affected (confirmed against a live grill — its frames arrive complete).

## Fix

`grills.json` is generated from the vendor API and must never be hand-edited (REVIEW.md), so the routine's own intended threshold `N` is recovered from its `parts < N` text and enforced in Python: `parse_status`/`parse_temperatures` return `None` for a message carrying fewer than `N` byte-pairs, before running the JS. It reconstructs exactly the guard the vendor wrote and failed to make effective. Boards whose guard already works are unaffected — a valid frame clears the same threshold.

## Tests

- `test_a_truncated_frame_is_rejected_not_parsed_into_a_falsy_state` — a header-only fragment on a dead-guard board parses to `None`, a full frame still parses.
- `test_every_board_accepts_a_frame_at_its_length_threshold` — no false drops: every board accepts a frame at its threshold and rejects one byte short.

Both fail against unpatched `main`, verified.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -290,6 +290,35 @@ def _live_code(js: str | None) -> str:
     return re.sub(r"//.*", "", re.sub(r"/\*.*?\*/", "", js, flags=re.DOTALL))
 
 
+def _min_frame_bytes(js: str | None) -> int | None:
+    """The byte count the routine's own guard requires, or None if it has no guard.
+
+    Every routine opens with `if (parts < N) return null` intending to reject
+    a frame too short to read -- but `parts` is the Array, and `Array < N`
+    coerces to `NaN < N`, always false, so the guard never fires. A truncated
+    frame then indexes past the end (every `parts[i]` is `undefined`), and the
+    routine returns a fully-populated, entirely-falsy state rather than
+    nothing. `parts` is the hex split into byte pairs, so the intended
+    threshold is `N` bytes; recovered here and enforced in `parse_*`, since
+    `grills.json` is generated and must not be hand-edited (REVIEW.md)."""
+    m = re.search(r"\bparts(?:\.length)?\s*<\s*(\d+)", _live_code(js))
+    return int(m.group(1)) if m else None
+
+
+def _is_truncated(message: str, js: str | None) -> bool:
+    """Whether `message` is shorter than the routine needs to read.
+
+    The vendor's dead `parts < N` guard, made to work: the routine splits the
+    hex into byte pairs, so a message carrying fewer than `N` bytes has it
+    reading past the end. A frame at or above the threshold is left alone.
+    """
+    min_bytes = _min_frame_bytes(js)
+    if min_bytes is None:
+        return False
+    # `parseHexMessage` steps by two chars, so the byte count is `ceil(len/2)`.
+    return (len(message) + 1) // 2 < min_bytes
+
+
 @dataclass(frozen=True)
 class ControlBoard:
     """Specifications for a control board connected via UART."""
@@ -367,6 +396,8 @@ class ControlBoard:
         """
         if not self._status_js_func:
             raise NotImplementedError
+        if _is_truncated(message, self._status_js_func):
+            return None
         return _drop_fields(
             self._evaljs(self._status_js_func, message),
             DROPPED_STATUS_FIELDS.get(self.name, frozenset()),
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Real bug, reproduced exactly as described. On `main`:

```
PBL.parse_status('FE0B00000000') -> 16 fields, all falsy
  moduleIsOn=False  noPellets=False  highTempErr=False
```

A fully-populated status dict that `_on_state_received` merges as truthy, clearing every error flag. That is the worst shape a parse bug can take — not a crash, not a dropped update, but a confident wrong answer that overwrites a correct one.

```
drop the status guard          -> test_a_truncated_frame_is_rejected_...   FAILS
drop the temps guard           -> test_a_truncated_frame_is_rejected_...   FAILS
off-by-one the other way (<=)  -> 13 tests FAIL
floor instead of ceil          -> 900 passed
```

## Two refinements to the description, neither affecting the fix

**The guard is not "always false".** I ran the coercions rather than reasoning about them:

```
[] < 10            -> True     empty array -> "" -> 0
['05'] < 10        -> True     one element -> "05" -> 5   (compares the VALUE, not the length)
['FE'] < 10        -> False    one element -> "FE" -> NaN
['FE','0B'] < 10   -> False    two or more -> comma -> NaN
```

So it fires for an empty array, and for a single byte it accidentally compares that byte's *value* against the threshold. It never fires from two bytes onward — which is every truncated frame that carries a header, so the conclusion and the reachability both stand. Worth stating precisely because "always false" invites someone to check it once, see `[] < 10` come back `True`, and doubt the whole finding.

**Counts, on reachable boards.** 14 of 20 is right for *declared* boards. Of the **17 reachable** ones it is **12 with the dead guard, 5 with the live one** — PBA, PBB, PBD, PBE, PBT. PBX1 is among the three unreachable, so it never sees a frame either way.

## The thing I checked hardest: does this drop anything real

The fix rejects a 44-byte status frame on PBA/PBB/PBD/PBE/PBT, whose thresholds are 47. That looked like a false drop until I ran `main` as a control — those five **already reject the same frame**, via the live `parts.length < 47` guard they have. Identical behaviour before and after, so it is their genuine frame length differing from PBL's, not a regression this introduces.

Thresholds recovered across the catalogue: status 36/41/44/47, temps 21/24/27/30 — all plausible frame lengths, no zeroes and no absurd values, which is the sanity check on parsing a number out of vendor source text.

## One gap

`floor` instead of `ceil` survives. It only differs on an odd-length message — exactly what a fragment truncated mid-byte-pair looks like — and `ceil` is the correct choice: `parseHexMessage` steps `i += 2`, so 87 chars yields 44 elements, not 43. Your comment says this; nothing enforces it. A single odd-length case at the threshold would pin it.

Not blocking — `floor` would only be over-strict, and only on a message that is malformed anyway.

## On the approach

Recovering the threshold from the routine's own text is the right call given `grills.json` is generated and must not be hand-edited. It is also the kind of thing that ages badly if the vendor changes the guard's wording, so the regex being anchored on `parts` (with or without `.length`) rather than on the whole statement is worth having — it keeps working if they fix their own bug.

PR #592, opened 2026-08-04
Read a disconnected probe as None, not -18 C

trailro:
## What

On the 9 converting boards (~82 supported models), a disconnected probe reads **-18 °C** instead of `None` when the grill is in Celsius.

The chain: `convertTemperature` (in the JS template) maps the disconnected-probe sentinel `960` to `null`. The board's `ftoc` helper then runs on that value, and its guard is `typeof temp === "undefined" || temp === 960` — it does **not** cover `null`. So `ftoc(null)` falls through to `Math.floor((null - 32) / 1.8)`; JS coerces `null` to `0`, giving `Math.floor(-17.7)` = **-18**. Every unplugged probe, target, and unused temperature field on a Celsius grill reads -18 °C.

Reproduced on all 9 boards; against unpatched `main`, PBA's sentinel frame yields `{p1Target: -18, p2Target: -18, p1Temp: -18, ... grillTemp: -18}`. Surfaced through `get_probe_targets`, `get_probe_targets`'s own docstring ("absent rather than present as None") is violated — it invents a -18 °C target for probes that have none. (A live PBL grill — a *non*-converting board — reads `None` correctly, and confirmed the two upstream premises: real grills send the 960 sentinel and do run in Celsius.)

## Fix

`grills.json` is generated from the vendor API and must never be hand-edited (REVIEW.md). Every `=== 960` in the shipped routines is that `ftoc` guard — nothing else compares to the sentinel — so `_scrub_js` (which already rewrites this JS at load time) extends the guard to `null` as well: `temp === 960 || temp === null`. `ftoc` returns its argument unchanged for the guarded cases, so `ftoc(null)` now returns `null` → the field is `None`, the intended value.

**Lossless:** the fix touches only the sentinel path. A genuine reading still converts (verified: 225 °F → 107 °C, unchanged); only `null` becomes `None` instead of -18. Non-converting boards never call `ftoc` and are entirely unaffected.

## Tests

- `test_a_disconnected_probe_reads_none_not_minus_18_on_converting_boards` — all 9 boards, every sentinel field, no -18 in the result.
- `test_a_real_temperature_still_converts_after_the_sentinel_fix` — a real value still converts; only the sentinel is None.

Both fail against unpatched `main`, verified.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -137,11 +137,23 @@ parse(dukpy['message']);
 
 _FN_RE = re.compile(r"(.+ ?= ?)(\(.[^\)]+\))( ?=>)?(.+)")
 
+# The converting boards' `ftoc` helper short-circuits the disconnected-probe
+# sentinel with `typeof temp === "undefined" || temp === 960`, returning the
+# value unchanged -- but `convertTemperature` (in the template) has already
+# mapped 960 to `null`, and the guard does not cover `null`. So `ftoc(null)`
+# falls through to `Math.floor((null - 32) / 1.8)`, and JS coerces `null` to
+# 0: a disconnected probe reads -18 (degrees C) instead of None. Every
+# `=== 960` in the shipped routines is that guard (nothing else compares to
+# the sentinel), so extending it to `null` restores the intended None without
+# touching `grills.json`, which is generated (REVIEW.md).
+_FTOC_SENTINEL_RE = re.compile(r"(\w+)\s*===\s*960\b")
+
 
 def _scrub_js(s: str | None) -> str | None:
     if s is None:
         return s
     s = _FN_RE.sub(r"\1 function \2\4", s)
+    s = _FTOC_SENTINEL_RE.sub(r"\1 === 960 || \1 === null", s)
     s = s.replace("let ", "var ")
     s = s.replace("const ", "var ")
     return s
--- a/tests/test_grills.py
+++ b/tests/test_grills.py
@@ -568,3 +568,40 @@ def test_every_board_accepts_a_frame_at_its_length_threshold():
                 continue
             assert not grills_lib._is_truncated("f" * (2 * n), js)  # at threshold: ok
             assert grills_lib._is_truncated("f" * (2 * n - 2), js)  # one byte short
+
+
+def test_a_disconnected_probe_reads_none_not_minus_18_on_converting_boards():
+    """`ftoc(null)` returned -18 for a disconnected probe on the ftoc boards.
+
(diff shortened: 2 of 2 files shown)


trailro:
Rebased onto `main` after #591 landed — the conflict was the tail of `tests/test_grills.py` (both PRs append tests), both kept. 906 pass, ruff/format/mypy clean.

dknowles2 (approved):
Verified, merging. **906 tests**, `ruff check`, `ruff format --check`, `mypy .` clean.

```
drop the sentinel rewrite                 -> test_a_disconnected_probe_reads_none_...   FAILS
rewrite to `undefined` instead of `null`  -> test_a_disconnected_probe_reads_none_...   FAILS
break it so real values are swallowed     -> 91 tests FAIL
```

That third one is the one I most wanted to see. The risk in this change is not that it fails to fix the sentinel, it is that it quietly stops converting real readings — and 91 failures says the catalogue-wide parse tests would catch that immediately.

## Checked the provenance claim by executing

"Every `=== 960` in the shipped routines is that guard" is the load-bearing assumption — a rewrite applied by regex to generated vendor source is only safe if you know every site it hits. Swept all reachable boards:

```
total `=== 960` occurrences: 17
distinct surrounding shapes: 1  (the ftoc guard)
```

No other comparison to the sentinel anywhere, so the substitution cannot land somewhere unintended.

Counts also check out on reachable boards: **9 converting boards** — LFS, PBA, PBE, PBL3, PBM, PBM2, PBT, PBV, PBV2 — carrying **82 models**, exactly as stated.

## One thing I nearly got wrong, and it is my error to record

I went to check that `_scrub_js` even reaches these routines, because I have been carrying a note saying it applies to *command* functions only and that status functions go to dukpy untouched. That note is wrong: `grills.py:357-358` runs both `status_function` and `temperature_function` through it. I had conflated `_scrub_js` with `_live_code` — the latter is the comment stripper, and the distinction is what my earlier note was actually about.

Worth saying out loud because it is the second time this session that a belief of mine about this file has been wrong in a way that would have made me reject a correct fix.

## On the shape of the fix

Extending the guard rather than special-casing `ftoc`'s callers is right, and the reason is in the JS: `ftoc` returns its argument unchanged for guarded values, so widening the guard makes `ftoc(null)` return `null` — the same path the sentinel already takes, not a new one. Nothing else in the function has to know.

The `-18` is a good tell for anyone reading a bug report later: `Math.floor((0 - 32) / 1.8)`. A probe reading exactly -18 °C on a Celsius grill is not cold, it is absent.

PR #593, opened 2026-08-04
Convert every temperature field a Celsius frame reports

trailro:
## What

PBA/PBE/PBT run their temperature reply through an `ftoc` helper when the grill is in Celsius — but the list of fields they convert omits `p4Temp` and `smokerActTemp`. Those two stay in Fahrenheit while every sibling field is Celsius: **two temperatures in one frame in different units**, with nothing marking which. Reproduced — a Celsius frame reading 212 °F at both grill and smoker:

```
PBA Celsius: grillTemp=100  smokerActTemp=212   <- 100 C vs a raw Fahrenheit 212
```

On a real Celsius grill `smokerActTemp` reads about twice too high. (`p4Temp` is the moot half — all 22 models have 2 probes, so probe 4 reads the disconnected sentinel → `None`; `smokerActTemp` is the live chamber reading and the real problem.)

## Fix

`grills.json` is generated and must not be hand-edited (REVIEW.md), so a per-board field set — the same shape `DROPPED_*_FIELDS` already uses for this class of routine gap — lists the fields the routine leaves behind, and `_convert_missed_fields` converts them with `ftoc`'s own arithmetic (`floor((f - 32) / 1.8)`) **only when the reply reports Celsius**, mirroring the board's own `if (!status.isFahrenheit)` block.

Verified safe:
- A Fahrenheit frame is untouched — no double conversion (`smokerActTemp` stays 212).
- The disconnected-probe sentinel is already `None` by then, so only real readings are converted.
- The affected set is exactly PBA/PBE/PBT (round-2 JS sweep confirmed every other converting board converts all the fields it populates); no other board is touched.

Note this is distinct from the `DROPPED` of `smokerActTemp` on LBL/LFS — there the field reads no bytes of its own (an overlap), here it's a real distinct reading in the wrong unit, so it's converted rather than dropped.

## Tests

- `test_smoker_temp_is_converted_to_celsius_like_its_siblings` — on all three boards, `smokerActTemp` now agrees with `grillTemp` (both 212 °F → 100 °C). Fails against unpatched `main`, verified.
- `test_a_fahrenheit_frame_leaves_the_missed_fields_untouched` — no double conversion on a Fahrenheit grill.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -1,9 +1,10 @@
 """Routines for accessing grill metadata."""
 
 import json
+import math
 import re
 import threading
-from collections.abc import Container, Iterable
+from collections.abc import Collection, Container, Iterable
 from dataclasses import dataclass, field
 from functools import cache
 from importlib import resources
@@ -90,6 +91,19 @@ DROPPED_STATUS_FIELDS = {
 DROPPED_TEMPERATURE_FIELDS = {
     "LBL": frozenset({"smokerActTemp"}),
     "LFS": frozenset({"smokerActTemp"}),
+}
+
+# Fields these boards' temperature routine populates but forgets to run
+# through its `ftoc` helper, so they stay in Fahrenheit while every sibling
+# field is converted to the grill's Celsius unit -- two temperatures in one
+# reply in different units, with nothing marking which. Converted here to
+# match, but only when the frame reports Celsius. `grills.json` is generated
+# and must not be hand-edited (REVIEW.md); `DROPPED_*_FIELDS` uses this same
+# per-board shape for the same class of routine gap.
+UNCONVERTED_FAHRENHEIT_FIELDS = {
+    "PBA": frozenset({"p4Temp", "smokerActTemp"}),
+    "PBE": frozenset({"p4Temp", "smokerActTemp"}),
+    "PBT": frozenset({"p4Temp", "smokerActTemp"}),
 }
 
 # Typos in the vendor's command slugs, mapped to the canonical slug.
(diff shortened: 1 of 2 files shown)


trailro:
Rebased onto `main` after #591 landed — real resolve in `parse_temperatures`: #591's truncation guard and this PR's `_convert_missed_fields` wrapper now compose (guard first, then parse+drop+convert), plus the tail-of-file test append. 906 pass, ruff/format/mypy clean.

trailro:
Rebased again onto `main` after #592 landed — same file, so it re-conflicted. Resolved by taking main's `tests/test_grills.py` (which now carries #592's tests) and re-appending only this PR's three finding-4 tests; the `grills.py` change applied cleanly. Teeth re-checked against main's grills.py (the smoker-conversion test fails without the fix). 908 pass, ruff/format/mypy clean.

dknowles2 (approved):
Verified, merging. **908 tests**, `ruff check`, `ruff format --check`, `mypy .` clean.

```
no-op the conversion                    -> test_smoker_temp_is_converted_to_celsius_...  FAILS
convert regardless of unit (double)     -> 23 tests FAIL
round instead of floor                  -> 908 passed
```

## I verified the affected set independently, before the rebase

"Exactly PBA/PBE/PBT" is the claim this whole change rests on, so I swept every converting board myself rather than taking the round-2 result:

```
LFS   ( 4 models)  7 temp fields,  7 converted,  MISSED []
PBA   ( 9 models)  9 temp fields,  7 converted,  MISSED ['p4Temp', 'smokerActTemp']
PBE   ( 2 models)  9 temp fields,  7 converted,  MISSED ['p4Temp', 'smokerActTemp']
PBL3  ( 4 models)  8 temp fields,  8 converted,  MISSED []
PBM   (13 models)  5 temp fields,  5 converted,  MISSED []
PBM2  ( 4 models)  5 temp fields,  5 converted,  MISSED []
PBT   (11 models)  9 temp fields,  7 converted,  MISSED ['p4Temp', 'smokerActTemp']
PBV   (30 models)  7 temp fields,  7 converted,  MISSED []
PBV2  ( 5 models)  7 temp fields,  7 converted,  MISSED []
```

Exactly three boards, exactly those two fields, 22 models. The set is right and it is complete.

## The gap: `floor` is load-bearing and nothing pins it

`round` instead of `math.floor` survives the suite. That is not a hypothetical difference:

```
Fahrenheit values in 32..500 where floor != round: 208 of 468
at 33 F: ftoc/floor -> 0 C, round -> 1 C
```

Nearly half of the practical range diverges. The reason the mutation survives is that `test_smoker_temp_is_converted_to_celsius_like_its_siblings` uses **212 °F, which is exactly 100 °C** — one of the values where floor and round agree, so it cannot tell the two implementations apart.

This matters more than a usual off-by-one, because it is the same finding as #327: these boards *truncate* where an obvious implementation rounds. With `round`, `smokerActTemp` would land one degree away from `grillTemp` for the same reading — the exact inconsistency this PR removes, reintroduced at 1-degree scale and much harder to notice than the 2× version.

**Not holding the merge** — the code is correct. One extra value in the existing parametrisation closes it; 213 °F (floor 100, round 101) would do.

## Two things I checked and am satisfied with

`is not False` rather than `== False` is the right guard: a frame with no `isFahrenheit` key at all is left alone rather than converted on a `None`. Conservative in the correct direction, since converting a Fahrenheit reading is the failure that would be worse than the bug.

The `isinstance(value, int) and not isinstance(value, bool)` guard survives mutation, but I am not raising it as a gap — the only fields in `UNCONVERTED_FAHRENHEIT_FIELDS` are `p4Temp` and `smokerActTemp`, neither of which is ever a bool, so there is no reachable path for it to matter. It is defensive against a future entry, which is a fine reason to keep it and a poor reason to write a test.

Distinguishing this from the `DROPPED` of `smokerActTemp` on LBL/LFS is the right call and worth the note you left: there the field reads no bytes of its own, here it is a real reading in the wrong unit. Same field name, opposite treatment, for reasons that only make sense with the routine in front of you.

PR #598, opened 2026-08-05
Drop PBVA's p4Temp, which reads the grill setpoint's bytes

trailro:
PBVA's temperature routine reads `p4Temp` and `grillSetTemp` from the same offset — bytes 14–16 of the FE0C frame:

```js
p4Temp: convertTemperature(parts, 14),
grillSetTemp: convertTemperature(parts, 14),
```

and the board's models (PBV30DS/PBV30DX) carry two meat probes, so `p4Temp` is not a probe at all: it reports the grill setpoint as a probe reading. The routine's Celsius conversion block also skips it, so on a Celsius grill the bogus value additionally stays in Fahrenheit next to a converted `grillSetTemp` — #593's class of gap, on a field that shouldn't exist in the first place.

The fix drops the field after parsing via `DROPPED_TEMPERATURE_FIELDS`, the same way LBL/LFS handle their aliased `smokerActTemp`, keeping the vendor routine byte-identical to `grills.json`. The status routine needs nothing: its whole temperature block is commented out, so it already emits no `p4Temp`.

Found by sweeping every board's routines with synthesized frames rather than the three boards named in the original report. Two reasons it survived #593:

- PBVA's models are in `UNSUPPORTED_MODELS`, so `all_variants()` never visits this board — no exhaustive test ever parsed its frames.
- The per-variant temperature tests carried a `p4Temp`/`smokerActTemp` tolerance ("Some grills don't convert these fields for some reason") that skipped exactly this class of mismatch. Every board those tests visit now converts correctly, so this PR also removes the tolerance in both tests — it's what kept the gap invisible.

The regression test goes through `get_grill("PBV30DS")`, the only path that reaches an unsupported board, and asserts the field is gone in both units while `grillSetTemp` still converts through the board's own `ftocSet` table (250 °F → 120 °C).

`uv run pytest` (914 passed), `ruff check`, `ruff format --check`, `mypy` all clean.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -91,6 +91,14 @@ DROPPED_STATUS_FIELDS = {
 DROPPED_TEMPERATURE_FIELDS = {
     "LBL": frozenset({"smokerActTemp"}),
     "LFS": frozenset({"smokerActTemp"}),
+    # PBVA's p4Temp reads no bytes of its own either: it duplicates
+    # grillSetTemp's offset in the FE0C frame, on a board whose models carry
+    # two meat probes -- so it reports the grill setpoint as a probe reading.
+    # The routine's conversion block also skips it, so on a Celsius grill the
+    # bogus reading additionally stays in Fahrenheit. Its status routine
+    # already emits no p4Temp (the whole temperature block there is commented
+    # out).
+    "PBVA": frozenset({"p4Temp"}),
 }
 
 # Fields these boards' temperature routine populates but forgets to run
--- a/tests/test_grills.py
+++ b/tests/test_grills.py
@@ -254,13 +254,7 @@ class TestGetGrills:
                 temp = want[key]
                 if js.converts_to_celsius():
                     temp = f_to_c(want[key])
-                try:
-                    assert status[key] == temp, f"{key}: {status[key]} != {temp}"  # type: ignore[literal-required]
-                except AssertionError:
-                    if key in ("p4Temp", "smokerActTemp"):
-                        # Some grills don't convert these fields for some reason.
-                        continue
-                    raise
+                assert status[key] == temp, f"{key}: {status[key]} != {temp}"  # type: ignore[literal-required]
 
     @pytest.mark.parametrize("grill", all_variants(), ids=idfn)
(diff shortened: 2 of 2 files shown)


dknowles2:
Verified the byte offsets independently rather than from the description — REVIEW.md records two board-remapping mistakes in this repo, so this is the class of claim that has to be checked against `grills.json` itself.

**Every claim holds.** From the PBVA `temperature_function`:

```js
p4Temp: convertTemperature(parts, 14),
grillSetTemp: convertTemperature(parts, 14),
```

Same offset, confirmed. And the conversion block covers everything except that field:

```
object literal assigns:  grillSetTemp, grillTemp, p1Temp, p2Temp, p3Temp, p4Temp
Celsius block converts:  grillSetTemp, grillTemp, p1Temp, p2Temp, p3Temp, p1Target, p2Target
```

`p4Temp` is the one field assigned and never converted — so on a Celsius grill it reported the setpoint in Fahrenheit next to a `grillSetTemp` converted through `ftocSet`. Both PBVA models carry `meat_probes=2`, and both are in `UNSUPPORTED_MODELS`, which is why `all_variants()` never reached this board.

**The test discriminates properly.** The frame is 22 bytes, which is exactly what the routine's `parts.length < 22` guard admits, and the Celsius assertion of 120 °C comes from the board's own `ftocSet` table rather than the generic `ftoc` — `floor((250 - 32) / 1.8)` would be 121, so a fix that routed the setpoint through the wrong converter would fail this test.

Three mutations, all caught:

```
drop the PBVA entry entirely                 caught
drop the wrong field (grillSetTemp)          caught
apply it to status instead of temperatures   caught
```

914 passed, ruff/format/mypy clean.

## On removing the tolerance

Worth stating plainly for the record, since it reads like part of the fix and isn't: removing the `p4Temp`/`smokerActTemp` `try/except` does **not** catch this bug, because those parametrized tests never visit PBVA. It stops the *next* one. The tolerance would have swallowed any future board whose conversion block skips a field, which is precisely how #593's class of gap hides — so removing it while every visited board converts correctly is the right moment to do it.

Merging.

PR #599, opened 2026-08-05
Add ControlBoard.emits() for asking what a board reports

trailro:
The boards do not all report the same state: PBM and PBM2 emit no `erL`, most boards no `primeState`, and a few probe fields exist only on some generations. A consumer creating an entity per field — ha-pitboss makes one binary sensor per error flag — needs to know which fields a board can ever produce, or the entity sits permanently unknown. Today the only way to ask is to parse a frame and see what comes back, which is exactly the wrong moment: entity creation happens before the first frame.

```py
>>> get_grill("PB1600PS1").control_board.emits("erL")
True
>>> get_grill("PB850CS2").control_board.emits("erL")   # PBM2
False
```

Design follows `converts_temperatures_to_celsius`: read off the routines' live code rather than maintaining a list of board names, so a definitions refresh keeps the answer current. Two refinements on top of plain substring membership, both borrowed from the test suite's `JSFunc.has_key`, where this predicate has been proven against every variant:

- **Key-assignment form** (`field:` in live code), not mere mention — LFS comments `p1Target` out of its status object literal but keeps a live `status.p1Target = ftoc(status.p1Target)`, which converts undefined and reports nothing.
- **Post-parse drops honored** — a field in `DROPPED_*_FIELDS` never reaches a consumer no matter what the routine assigns: LBL's `smokerActTemp` is assigned in both routines and reported by neither, while its `grillTemp` (dropped from status only) still arrives via the temperatures reply. Both are asserted in the tests.

Semantics are "can report", not "always reports" — `grillSetTemp` appears only while the grill holds a setpoint, and that's still `True`: the question consumers ask is whether the field is worth an entity.

Downstream in ha-pitboss this gates the "Startup error" binary sensor (and lets similar per-board gaps be handled without a hardcoded board list), keeping the board-knowledge boundary where REVIEW.md draws it: facts about boards live here, consumers just ask.

`uv run pytest` (916 passed), `ruff check`, `ruff format --check`, `mypy` all clean.

### DIFF
--- a/pytboss/grills.py
+++ b/pytboss/grills.py
@@ -438,6 +438,37 @@ class ControlBoard:
         """
         return "ftoc" in _live_code(self._status_js_func)
 
+    def emits(self, field: str) -> bool:
+        """Whether this board's routines can report `field`.
+
+        The boards do not all report the same state: PBM and PBM2 emit no
+        `erL`, most emit no `primeState`, and a few probe fields exist only
+        on some generations. A consumer creating an entity per field needs
+        to know which fields this board can ever produce, or the entity sits
+        permanently unknown.
+
+        True when live code in either parsing routine assigns the key in the
+        object it returns, and the field survives the post-parse drops
+        (`DROPPED_*_FIELDS`). Read off the routines rather than a list of
+        board names, like `converts_temperatures_to_celsius`: a definitions
+        refresh can add or remove a field, and only live code counts. The
+        key-assignment form matters too -- LFS comments `p1Target` out of
+        its status object literal but keeps a live
+        `status.p1Target = ftoc(status.p1Target)`, which converts undefined
+        and reports nothing.
+
+        "Can report", not "always reports": some fields are conditional,
+        like `grillSetTemp` appearing only while the grill holds a setpoint.
+        """
+        key = re.compile(rf"\b{re.escape(field)}\s*:")
+        if field not in DROPPED_STATUS_FIELDS.get(
+            self.name, frozenset()
+        ) and key.search(_live_code(self._status_js_func)):
+            return True
+        return field not in DROPPED_TEMPERATURE_FIELDS.get(
+            self.name, frozenset()
+        ) and bool(key.search(_live_code(self._temperatures_js_func)))
+
     def parse_status(self, message: str) -> StateDict | None:
         """Parses a status message.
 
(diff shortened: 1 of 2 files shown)


dknowles2:
The design argument is right — reading the routines beats a board list, for the same reason `converts_temperatures_to_celsius` does — so I spent the review on the one thing that would actually hurt a consumer: **`emits()` returning `False` for a field the board really reports**. That is the direction that makes ha-pitboss skip an entity the grill populates, and it is silent.

**Swept every board, both routines, both units — 880 field observations, zero false negatives.** Every field that appeared in a parsed reply has `emits() is True`.

That sweep also surfaced the case I was most worried about, which is the inverse of the LFS one you documented: fields set as `obj.field = ...` with no `field:` literal, where the regex would miss them.

```
LBL   status grillSetTemp    emits()=True
LBL   status grillTemp       emits()=True
LFS   status grillSetTemp    emits()=True
LFS   status grillTemp       emits()=True
LFS   status p1Target        emits()=False
LFS   temps  p1Target        emits()=False
PBV   temps  p1Target        emits()=False
PBV2  temps  p1Target        emits()=False
```

The `True`s are right for a reason worth writing down: LBL/LFS drop `grillSetTemp` and `grillTemp` from *status*, and `emits()` answers `True` off the **temperatures** routine's literal — which is exactly where those values really arrive. The drop table and the two-routine check work together there rather than either one alone getting it right.

The `False`s I checked by parsing rather than reasoning, since `temps.p1Target = ftoc(temps.p1Target)` assigning `undefined` could plausibly have created a key:

```
PBV   F/C   p1Target in reply: False
PBV2  F/C   p1Target in reply: False
LFS   F/C   p1Target in reply: False
```

It does not survive to the reply, in either unit. `False` is correct.

**Mutations, all caught:**

```
key-assignment form -> bare mention      caught
read raw source instead of live code     caught
ignore the status drops                  caught
ignore the temperature drops             caught
only look at the status routine          caught
```

**Composed with #598**, which landed first: 917 passed, ruff/format/mypy clean, and the sweep re-run on the merged tree still shows zero false negatives — PBVA's new `p4Temp` drop flows through `emits()` for free, which is the point of reading `DROPPED_*_FIELDS` rather than restating it.

Merging. I see #388 for the ha-pitboss half.

